PR #1, opened 2016-06-29
CENSURE Keypoint Detector

mronian:
- [x] DataTypes
- [x] Filters
  - [x] Box
  - [x] Octagon
  - [ ] Star
- [x] Filter Response
  - [x] Box
  - [x] Octagon
  - [ ] Star
- [x] Feature Detect

### DIFF
--- a/src/censure.jl
+++ b/src/censure.jl
@@ -14,11 +14,14 @@ end
 
 type OctagonFilter <: BiFilter
 	
-	filter :: Array{Float64, 2}
-	m_out :: UInt8
-	m_in :: UInt8
-	n_out :: UInt8
-	n_in :: UInt8
+	m_out :: Integer
+	m_in :: Integer
+	n_out :: Integer
+	n_in :: Integer
+	in_area :: Float64
+	out_area :: Float64
+	in_weight :: Float64
+	out_weight :: Float64
 
 end
 
@@ -33,33 +36,21 @@ type CENSURE <: Detector
 
 end
 
-function createFilter(OF::OctagonFilter)
-	area_out = (OF.m_out + 2 * OF.n_out) ^ 2 - 2 * OF.n_out ^ 2
-	area_in = (OF.m_in + 2 * OF.n_in) ^ 2 - 2 * OF.n_in ^ 2
-	weight_out = 1.0/(area_out - area_in)
-	weight_in = 1.0/area_in
-	OF.filter[:, :] = weight_out
-	inner_start = Int(0.5 * ((OF.m_out + 2 * OF.n_out) - (OF.m_in + 2 * OF.n_in)))
-	OF.filter[inner_start + 1 : end - inner_start, inner_start + 1 : end - inner_start] = weight_in
-
-	for i in 1:OF.n_in
-		OF.filter[inner_start + i, inner_start + 1 : inner_start + OF.n_in - i + 1] = weight_out
-		OF.filter[inner_start + i, inner_start + OF.n_in + OF.m_in + i : inner_start + OF.m_in + 2 * OF.n_in] = weight_out
-		OF.filter[inner_start + OF.m_in + 2 * OF.n_in - i + 1, inner_start + 1 : inner_start + OF.n_in - i + 1] = weight_out
-		OF.filter[inner_start + OF.m_in + 2 * OF.n_in - i + 1, inner_start + OF.n_in + OF.m_in + i : inner_start + OF.m_in + 2 * OF.n_in] = weight_out
-	end
-
-	for i in 1:OF.n_out
-		OF.filter[i, 1 : OF.n_out - i + 1] = 0
-		OF.filter[i, OF.n_out + OF.m_out + i : OF.m_out + 2 * OF.n_out] = 0
-		OF.filter[OF.m_out + 2 * OF.n_out - i + 1, 1 : OF.n_out - i + 1] = 0
-		OF.filter[OF.m_out + 2 * OF.n_out - i + 1, OF.n_out + OF.m_out + i : OF.m_out + 2 * OF.n_out] = 0
-	end
-
+function OctagonFilter(mo, mi, no, ni)
+	OF = OctagonFilter(mo, mi, no, ni, 0.0, 0.0, 0.0, 0.0)
+	OF.out_area = OF.m_out ^ 2 + 2 * OF.n_out ^ 2 + 4 * OF.m_out * OF.n_out
+	OF.in_area = OF.m_in ^ 2 + 2 * OF.n_in ^ 2 + 4 * OF.m_in * OF.n_in
+	OF.out_weight = 1.0 / (OF.out_area - OF.in_area)
+	OF.in_weight = 1.0 / OF.in_area
+	OF
 end
 
-OctagonFilter(mo, mi, no, ni) = (OF = OctagonFilter(ones(Float64, mo + 2 * no, mo + 2 * no), mo, mi, no, ni); createFilter(OF); OF)
-BoxFilter(s) = (BF = BoxFilter(s, 2 * s + 1, 4 * s + 1, (2 * s + 1) ^ 2, (4 * s + 1) ^ 2, 0.0, 0.0); BF.in_weight = 1.0 / BF.in_area; BF.out_weight = 1.0 / (BF.out_area - BF.in_area); BF; )
+function BoxFilter(s)
+	BF = BoxFilter(s, 2 * s + 1, 4 * s + 1, (2 * s + 1) ^ 2, (4 * s + 1) ^ 2, 0.0, 0.0)
+	BF.in_weight = 1.0 / BF.in_area
+	BF.out_weight = 1.0 / (BF.out_area - BF.in_area)
+	BF
+end
 
 OctagonFilter_Kernels = [
 						[5, 3, 2, 0],
@@ -78,47 +69,6 @@ Kernels = Dict(BoxFilter => BoxFilter_Kernels, OctagonFilter => OctagonFilter_Ke
 function getFilterStack(filter_type::Type, smallest::Integer, largest::Integer)
 	k = Kernels[filter_type]
 	filter_stack = map(f -> filter_type(f...), k[smallest : largest])
-end
-
-function filterResponse{T}(int_img::AbstractArray{T, 2}, filter::BoxFilter)
-	margin = filter.scale * 2
-	n = filter.scale
-	img_shape = size(int_img)
-	response = zeros(img_shape)
-	R = CartesianRange(CartesianIndex((margin + 1, margin + 1)), CartesianIndex((img_shape[1] - margin, img_shape[2] - margin))) 
-    in_sum = 0.0
-    out_sum = 0.0
-    for I in R
-    	topleft = I + CartesianIndex(- n - 1, - n - 1)
-    	topright = I + CartesianIndex(n, - n - 1)
-    	bottomleft = I + CartesianIndex(- n - 1, n)
-    	bottomright = I + CartesianIndex(n, n)
-    	A = topleft >= CartesianIndex(1, 1) ? int_img[topleft] : 0.0
-    	B = topright >= CartesianIndex(1, 1) ? int_img[topright] : 0.0
-    	C = bottomleft >= CartesianIndex(1, 1) ? int_img[bottomleft] : 0.0
-    	D = bottomright >= CartesianIndex(1, 1) ? int_img[bottomright] : 0.0
-    	in_sum = A + D - B - C
-
-    	topleft = I + CartesianIndex(- 2 * n - 1, - 2 * n - 1)
-    	topright = I + CartesianIndex(2 * n, - 2 * n - 1)
-    	bottomleft = I + CartesianIndex(- 2 * n - 1, 2 * n)
-    	bottomright = I + CartesianIndex(2 * n, 2 * n)
-    	A = topleft >= CartesianIndex(1, 1) ? int_img[topleft] : 0.0
-    	B = topright >= CartesianIndex(1, 1) ? int_img[topright] : 0.0
-    	C = bottomleft >= CartesianIndex(1, 1) ? int_img[bottomleft] : 0.0
-    	D = bottomright >= CartesianIndex(1, 1) ? int_img[bottomright] : 0.0
-    	out_sum = A + D - B - C - in_sum
-
-    	response[I] = in_sum * filter.in_weight - out_sum * filer.out_weight
-    end
-
-    response
-end
-
-function filterResponse(int_imgs::Tuple, filter::OctagonFilter)
-	int_img = int_imgs[1]
-	rs_img = int_imgs[2]
-	ls_img = int_imgs[3]
 end
 
 getIntegralImage(img, filter_type::BoxFilter) = integral_image(img)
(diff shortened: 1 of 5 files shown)


src/censure.jl
```diff
@@ -0,0 +1,223 @@
...
+	response = zeros(img_shape)
+	R = CartesianRange(CartesianIndex((margin + 1, margin + 1)), CartesianIndex((img_shape[1] - margin, img_shape[2] - margin))) 
+    in_sum = 0.0
+    out_sum = 0.0
+    for I in R
+    	topleft = I + CartesianIndex(- n - 1, - n - 1)
+    	topright = I + CartesianIndex(- n - 1, n)
+    	bottomleft = I + CartesianIndex(n, - n - 1)
+    	bottomright = I + CartesianIndex(n, n)
+    	A = topleft >= CartesianIndex(1, 1) ? int_img[topleft] : 0.0
```
mronian: Whats the correct way to check bounds on CartesianIndex?

``` julia
julia> a = CartesianIndex(2, 0)
CartesianIndex{2}((2,0))

julia> b = CartesianIndex(0, 2)
CartesianIndex{2}((0,2))

julia> a >= CartesianIndex(1, 1)
false

julia> b >= CartesianIndex(1, 1)
true
```

I need both the cases to be false for this to work correctly.

src/censure.jl
```diff
@@ -0,0 +1,223 @@
...
+
+function getFilterStack(filter_type::Type, smallest::Integer, largest::Integer)
+	k = Kernels[filter_type]
+	filter_stack = map(f -> filter_type(f...), k[smallest : largest])
+end
+
+getIntegralImage(img, filter_type::BoxFilter) = integral_image(img)
+
+function getIntegralImage(img, filter_type::OctagonFilter)
+	img_shape = size(img)
```
mronian: This function computes the integral and slant-integral images (right_slant holds the sum of pixels to the top and left as well as the pixels to the top and right till an angle of 45 degrees whereas left_slant holds the sum of pixels to the top and left till an angle of 45 degrees.)

Here the functions go in a row major fashion. I wanted to use CartesianRange but it goes in a coloumn major fashion which is the way julia stores arrays. Does the `Image` type store data in a row major fashion? I was a bit confused after going through the documentation.

I could transpose the image matrix and then take the integral images which will be faster if it is in a coloumn major fashion. But transposing just gives a view of the matrix right? Does it actually transpose the matrix in memory?

mronian:
@timholy I have some doubts which I have written inline as comments on the files. I have tried to make the code fast by looking at benchmarks and ProfileView(Thanks for that!) checked the type warning for most things.

src/censure.jl
mronian: Found `checkbounds()`. I'll make the changes :)

src/censure.jl
timholy: Images is very confused about orientation: I tried to put that info into the Dict (`spatialorder`). However, that has not proved to be a viable solution because taking `data` on the image throws that information away (as does `img[:,:]`). The upcoming ImagesCore will build that into the type system, by leveraging `PermutedDimsArray` (new to julia-0.5) to make it easy to flip the orientation lazily. I haven't yet decided whether we should make row-major or col-major the default (each has its advantages...). EDIT: input is welcome.

Taking the transpose _does_ transpose in memory, but the PermutedDimsArray will be a view.

src/censure.jl
```diff
@@ -0,0 +1,236 @@
...
+    	response[I] = in_sum * filter.in_weight - out_sum * filer.out_weight
+    end
+	response
+	
+end
+
+function checkFeature()
+end
+
+CENSURE(; smallest::Integer = 1, largest::Integer = 7, filter::Type = BoxFilter, responseThreshold::Number = 0.15, lineThreshold::Number = 10) = CENSURE(smallest, largest, filter, getFilterStack(filter, smallest, largest), responseThreshold, lineThreshold)
```
timholy: BTW, can you wrap long lines? It means I have to scroll my browser window to find the "comment" button :smile:.

src/censure.jl
```diff
@@ -0,0 +1,236 @@
...
+						[7, 3, 3, 2],
+						[9, 5, 4, 2],
+						[9, 5, 7, 3],
+						[13, 5, 7, 4],
+						[15, 5, 10, 5]
+						]
+
+BoxFilter_Kernels = [1, 2, 3, 4, 5, 6, 7]
+
+Kernels = Dict(BoxFilter => BoxFilter_Kernels, OctagonFilter => OctagonFilter_Kernels)
```
timholy: While I'm thinking of it (for future reference): when you get to the point of having this fully functional and want to turn on package precompilation, you'll need to `rehash` this Dict in the module's `__init__` function. That's because it's a Dict of types. More info at http://docs.julialang.org/en/latest/manual/modules/?highlight=__init__#module-initialization-and-precompilation.

timholy: An alternative to a dict would simply be a function

``` jl
getkernel(::Type{BoxFilter}) = BoxFilter_Kernels
getkernel(::Type{OctagonFilter}) = OctagonFilter_Kernels
```

This would be type-stable, unlike the Dict solution.

src/censure.jl
```diff
@@ -0,0 +1,236 @@
...
+end
+
+function BoxFilter(s)
+	BF = BoxFilter(s, 2 * s + 1, 4 * s + 1, (2 * s + 1) ^ 2, (4 * s + 1) ^ 2, 0.0, 0.0)
+	BF.in_weight = 1.0 / BF.in_area
+	BF.out_weight = 1.0 / (BF.out_area - BF.in_area)
+	BF
+end
+
+OctagonFilter_Kernels = [
```
timholy: add `const` in front of this

src/censure.jl
```diff
@@ -0,0 +1,236 @@
+abstract BiFilter
+
+type BoxFilter <: BiFilter
+
+	scale :: Integer
```
timholy: Only just noticed now, but there are a few indentation issues in this file. Not a big deal, but FYI 4 spaces is the convention.

src/censure.jl
```diff
@@ -0,0 +1,236 @@
...
+	in_length :: Integer
```
timholy: This is going to hurt performance because `Integer` is an abstract type. Better to make it an `Int` or, if you need flexibility about the type, make `BoxFilter` parametric, e.g.,

``` jl
type BoxFilter{T<:Integer} <: BiFilter
    scale::T
    out_length::T
    in_area::Float64
...
end
```

src/censure.jl
```diff
@@ -0,0 +1,236 @@
...
+	out_weight :: Float64
+
+end
+
+type CENSURE <: Detector
+
+	smallest :: Integer
+	largest :: Integer
+	filter_type :: Type
+	filter_stack :: Array
```
timholy: `Array`, `Type`, and `Number` are also abstract. Use parametric types where necessary, e.g.:

``` jl
type CENSURE{F,T,N} <: Detector
    smallest::Int
    largest::Int
    filter_type::F
    filter_stack::Array{T,N}
    responseThreshold::T
    lineThreshold::T
end
```

(There's a curious mix of underscores and camelcase here.)

src/censure.jl
```diff
@@ -0,0 +1,236 @@
...
+	R = CartesianRange(CartesianIndex((margin + 1, margin + 1)), CartesianIndex((img_shape[1] - margin, img_shape[2] - margin))) 
+    in_sum = 0.0
+    out_sum = 0.0
+    for I in R
+    	topleft = I + CartesianIndex(- n - 1, - n - 1)
+    	topright = I + CartesianIndex(- n - 1, n)
+    	bottomleft = I + CartesianIndex(n, - n - 1)
+    	bottomright = I + CartesianIndex(n, n)
+    	A = topleft >= CartesianIndex(1, 1) ? int_img[topleft] : 0.0
+    	B = topright >= CartesianIndex(1, 1) ? int_img[topright] : 0.0
```
timholy: To save typing you can assign consts before the loop, e.g. `o = one(CartesianIndex{2})` and just use `o`.

src/censure.jl
```diff
@@ -0,0 +1,236 @@
...
+	img_shape = size(int_img)
+	response = zeros(img_shape)
+	R = CartesianRange(CartesianIndex((margin + 1, margin + 1)), CartesianIndex((img_shape[1] - margin, img_shape[2] - margin))) 
+    
+    for I in R
+    	topleft = I + CartesianIndex(- m_in2 - 1, - m_in2 - OF.n_in - 1)
+    	topright = I + CartesianIndex(m_in2, - m_in2 - OF.n_in - 1)
+    	bottomleft = I + CartesianIndex(- m_in2 - 1, m_in2 + OF.n_in)
+    	bottomright = I + CartesianIndex(m_in2, m_in2 + OF.n_in)
+    	A = topleft >= CartesianIndex(1, 1) ? int_img[topleft] : 0.0
```
timholy: You may know you can also use logic like `if 1 < I[1] < n`...

src/censure.jl
timholy: Note that this only applies to types in fields, not arguments in functions (there is no penalty for declaring an argument as `dim::Integer`). The reason is that a type-declaration is _prospective_ information for use in inference; in contrast, method specialization uses the actual types of arguments when called.

mronian:
@timholy This is ready for review :)

src/censure.jl
```diff
@@ -0,0 +1,247 @@
...
+const octagon_filter_kernels = [
+						[5, 3, 2, 0],
+						[5, 3, 3, 1],
+						[7, 3, 3, 2],
+						[9, 5, 4, 2],
+						[9, 5, 7, 3],
+						[13, 5, 7, 4],
+						[15, 5, 10, 5]
+						]
+
```
mronian: Will fix the indentation.

src/censure.jl
```diff
@@ -0,0 +1,247 @@
...
+end
+
+function CENSURE(; smallest::Integer = 1, largest::Integer = 7, filter::Type = BoxFilter, response_threshold::Number = 0.15, line_threshold::Number = 10)
+    CENSURE{filter}(smallest, largest, filter, _get_filter_stack(filter, smallest, largest), response_threshold, line_threshold)
+end
+
+function censure{T, F}(img::AbstractArray{T, 2}, params::CENSURE{F})
+	int_img = _get_integral_image(img, params.filter_stack[1])
+	responses = map(f -> _filter_response(int_img, f), params.filter_stack)
+    response_matrix = reshape(hcat(responses...), size(img)..., size(responses)...)
```
mronian: Any faster way of doing this?

src/censure.jl
```diff
@@ -0,0 +1,247 @@
...
+    minima, maxima = extrema_filter(convert(Array{Float64}, padarray(response_matrix, [1, 1, 1], [1, 1, 1], "replicate")), [3, 3, 3])
+    features = map(i -> (minima[i] == response_matrix[i] || maxima[i] == response_matrix[i]) && ( response_matrix[i] > params.response_threshold ), CartesianRange(size(response_matrix)))
```
mronian: @timholy I ran `@profile` on this and `_filter_response` and `extrema_filter` are the two steps taking a lot of time. The memory allocated is huge too (~1.7 GB for the mandrill image ). Any suggestions which might help reduce this?

mronian: I ran it line by line and `extrema_filter` takes 90% of the total time and around 60% of memory allocations, `_filter_response` takes the remaining.

mronian: I've made some changes to `_filter_response` and amended the commit. Now `extrema_filter` accounts for 90% of both time and memory.

timholy: What's the array size? (Looks like it's an `Array{64,3}`, right?)

Looks like `extrema_filter` allocates temporary storage frequently, but I haven't looked at how it works all that carefully. It might be possible to speed it up by reusing the same buffer many times.

mronian:
@timholy I am changing the Keypoint type to be an alias of CartesianIndex{2} instead of the current type since it is easier for calculating descriptors (can be seen in the upcoming BRIEF descriptor) -

``` jl
type Keypoint
    x::Integer
    y::Integer
end 
```

Do you have a better representation in mind?

timholy:
> I am changing the Keypoint type to be an alias of CartesianIndex{2} instead of the current type

That's a good idea---you'll get a lot of very nice behavior for free that way.

For future reference when you do create your own types, remember that `Integer` is an abstract type and that you should avoid using abstract types as fields in a type if at all possible because performance will be terrible. It's much better to use a parametric type,

``` jl
type KeyPoint{T<:Integer}
    x::T
    y::T
end
```

timholy:
Refs:
http://docs.julialang.org/en/latest/manual/performance-tips/#avoid-containers-with-abstract-type-parameters
http://docs.julialang.org/en/latest/manual/performance-tips/#avoid-fields-with-abstract-type
http://docs.julialang.org/en/latest/manual/performance-tips/#avoid-fields-with-abstract-containers

src/censure.jl
```diff
@@ -0,0 +1,246 @@
...
+    for i in 1:params.largest - params.smallest + 1
+        gamma = (1 + (params.smallest + i - 1) / 3.0)
+        filt_cov_xx = imfilter_gaussian(cov_xx, [gamma, gamma])
+        filt_cov_xy = imfilter_gaussian(cov_xy, [gamma, gamma])
+        filt_cov_yy = imfilter_gaussian(cov_yy, [gamma, gamma])
+
+        features[:, :, i] = map((xx, yy, xy, f) -> (xx + yy) ^ 2 > params.line_threshold * (xx * yy - xy ^ 2) ? false : f, filt_cov_xx, filt_cov_yy, filt_cov_xy, features[:, :, i])
+    end
+    keypoints = Array{Keypoint}([])
+    scales = Array{Integer}([])
```
timholy, line 239: abstract container type. How about just `Array{Int}(0)`?

timholy:
Aside from very small points, this seems fine to me. (I'll have to play with it to really kick the tires, but it looks very promising!)

tejus-gupta:
@mronian You don't seem to be working on this. Can I finish it off?

mronian:
Yeah sure. It is mostly ready, you will need to update the code to the latest Julia. I did not merge it since the docs were not there.

tejus-gupta:
In filter response for octagon filter,
```
topright = I + CartesianIndex(m_in2, - m_in2 - OF.n_in - 1)
bottomleft = I + CartesianIndex(- m_in2 - 1, m_in2 + OF.n_in)
```
I don't think this is correct. The error wouldn't effect the filter response though.

mronian:
The filter response was correct since I had manually checked it for a few octagons and images. You can check it again to confirm. What seems to be the error in the code?

tejus-gupta:
It should be
topright = I + CartesianIndex(- m_in2 - 1, m_in2 + OF.n_in)
bottomleft = I + CartesianIndex(m_in2, - m_in2 - OF.n_in - 1)

The filter response would come out correctly because in_sum=A + D - B - C. Swapping B and C wouldn't change in_sum.

mronian:
@tejus-gupta I don't think its wrong. `topright` has to be obtained by decreasing the y-coordinate and increasing the x-coordinate and for `bottomleft` you will have to increase the y and decrease the x from the centre I.

mronian:
Whoops. I mistook the lines you mentioned as those from the code instead of corrections. Agreed, it is incorrect. Go ahead, @tejus-gupta 😄 Sorry for the trouble!

PR #2, opened 2016-07-06
Adds GLCM Function

mronian:
@timholy Some of the functions for glcm_prop are named differently in different sources. Would it be better to add wrapper functions for those which then call the correct one? (Eg. ASM -> Homogeneity, Contrast -> Inertia etc)

### DIFF
--- a/src/ImageFeatures.jl
+++ b/src/ImageFeatures.jl
@@ -6,8 +6,6 @@ using Images, ColorTypes, FixedPointNumbers
 
 include("glcm.jl")
 
-import ColorTypes: U8, U16, Gray
-
 export 
 	glcm,
 	glcm_symmetric,
--- a/src/ImageFeatures.jl
+++ b/src/ImageFeatures.jl
@@ -2,8 +2,13 @@ module ImageFeatures
 
 # package code goes here
 
-using Images
+using Images, ColorTypes, FixedPointNumbers
 
 include("glcm.jl")
+
+import ColorTypes: U8, U16, Gray
+
+export 
+	glcm
 
 end # module
--- a/src/ImageFeatures.jl
+++ b/src/ImageFeatures.jl
@@ -2,4 +2,8 @@ module ImageFeatures
 
 # package code goes here
 
+using Images
+
+include("glcm.jl")
+
 end # module
--- a/src/ImageFeatures.jl
+++ b/src/ImageFeatures.jl
@@ -9,6 +9,22 @@ include("glcm.jl")
 import ColorTypes: U8, U16, Gray
 
 export 
-	glcm
+	glcm,
+	glcm_symmetric,
+	glcm_norm,
+	glcm_prop,
+	max_prob,
+	contrast,
+	ASM,
+	IDM,
+	entropy,
+	energy,
+	contrast,
+	dissimilarity,
+	correlation,
+	glcm_mean_ref,
+	glcm_mean_neighbour,
+	glcm_var_ref,
+	glcm_var_neighbour
 
 end # module
--- a/src/ImageFeatures.jl
+++ b/src/ImageFeatures.jl
@@ -17,7 +17,7 @@ export
 	contrast,
 	ASM,
 	IDM,
-	entropy,
+	glcm_entropy,
 	energy,
 	contrast,
 	dissimilarity,
--- a/src/glcm.jl
+++ b/src/glcm.jl
@@ -1,154 +1,150 @@
 function glcm{T<:Real}(img::AbstractArray{T, 2}, distance::Integer, angle::Real, mat_size::Integer = 16)
-	max_img = maximum(img)
-	img_rescaled = map(i -> max(1, Int(ceil(i * mat_size / max_img))), img)
-	_glcm(img_rescaled, distance, angle, mat_size)
+    max_img = maximum(img)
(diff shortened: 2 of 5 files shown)


src/glcm.jl
```diff
@@ -0,0 +1,154 @@
...
+glcm{T<:Union{Colorant, Real}}(img::AbstractArray{T, 2}, distances::Array{Int, 1}, angles::Real, mat_size::Integer = 16) = glcm(img, distances, [angles], mat_size)
+
+function _glcm{T}(img::AbstractArray{T, 2}, distance::Integer, angle::Number, mat_size::Integer)
+	co_oc_matrix = zeros(Integer, mat_size, mat_size)
+
+	R = CartesianRange(size(img))
+
+	for I in R
+
+	int_one = img[I]
```
timholy: indentation

src/glcm.jl
```diff
@@ -0,0 +1,154 @@
...
+
+	co_occuring_pixel = I + CartesianIndex(Int(round(sin(angle) * distance)), Int(round(cos(angle) * distance)))
```
timholy: You're not getting a lot of benefit from `CartesianIndex` here because your algorithm is explicitly two-dimensional. If it only makes sense as a 2d algorithm, then it might be easier to just write this as `for j = 1:size(img,2), i = 1:size(img,1)`...

timholy:
> Some of the functions for glcm_prop are named differently in different sources. Would it be better to add wrapper functions for those which then call the correct one? (Eg. ASM -> Homogeneity, Contrast -> Inertia etc)

I would tend to pick just one, but I am not sufficiently aware of these issues to have a strong opinion.

mronian:
Added tests. @timholy This is ready for review.

src/ImageFeatures.jl
```diff
@@ -2,4 +2,29 @@ module ImageFeatures
 
 # package code goes here
 
+using Images, ColorTypes, FixedPointNumbers
+
+include("glcm.jl")
+
+import ColorTypes: U8, U16, Gray
```
timholy: Can't you just say `using ColorTypes`? This line suggests your actual intent is to _extend_ `U8`, `U16`, and `Gray`. (`import` allows you to extend them, whereas `using` does not).

src/glcm.jl
```diff
@@ -0,0 +1,150 @@
...
+    img_rescaled = map(i -> max(1, Int(ceil((convert(Gray{U8}, i).val.i) * mat_size / 256))), img)  
+    glcm_matrices = [_glcm(img_rescaled, d, a, mat_size) for d in distances, a in angles]
+    glcm_matrices
+end
+
+glcm{T<:Union{Colorant, Real}, A<:Real}(img::AbstractArray{T, 2}, distances::Int, angles::Array{A, 1}, mat_size::Integer = 16) = glcm(img, [distances], angles, mat_size)
+glcm{T<:Union{Colorant, Real}}(img::AbstractArray{T, 2}, distances::Array{Int, 1}, angles::Real, mat_size::Integer = 16) = glcm(img, distances, [angles], mat_size)
+
+function _glcm{T}(img::AbstractArray{T, 2}, distance::Integer, angle::Number, mat_size::Integer)
+    co_oc_matrix = zeros(Integer, mat_size, mat_size)
```
timholy: Abstract eltype, will be very slow.

src/glcm.jl
```diff
@@ -0,0 +1,150 @@
...
+    for j = 1:size(img, 1), i = 1:size(img, 2)
+        int_one = img[j, i]
+        j_off = j + Int(round(sin(angle) * distance))
```
timholy: precompute `sin(angle)` and `cos(angle)` and store them as constants---they're fairly slow, and since they are const there is no point recomputing them on each iteration of the loop.

If you really want to get fancy, you can even break this into two loops and do the bounds-checking on `j` and perhaps skip the entire loop over `i`...

src/glcm.jl
```diff
@@ -0,0 +1,150 @@
...
+function ASM{T<:Real}(glcm_window::Array{T, 2})
+    sum(map(i -> i ^ 2, glcm_window))   
+end
+
+function IDM{T<:Real}(glcm_window::Array{T, 2})
+    sum([glcm_window[id] / (1 + (id[1] - id[2]) ^ 2) for id in CartesianRange(size(glcm_window))])
+end
+
+function glcm_mean_ref{T<:Real}(glcm_window::Array{T, 2})
+    sumref = mapslices(sum, glcm_window, 2)
```
timholy: Unfortunately, as of now it's considerably faster to use `sum(glcm_window, 2)`.

test/glcm.jl
```diff
@@ -0,0 +1,121 @@
...
+        expected = [  8.0  13.0  18.0  23.0  28.0  28.0
+                      9.0  14.0  19.0  24.0  29.0  29.0
+                     10.0  15.0  20.0  25.0  30.0  30.0
+                     10.0  15.0  20.0  25.0  30.0  30.0
+                     10.0  15.0  20.0  25.0  30.0  30.0 ]
+        @fact all(glcm_props .== expected) --> true
+
+    end
+
+end
```
timholy, line 138: :+1: on your tests!

mronian:
Added more tests to aim for 100% coverage. @timholy Shall I merge this?

timholy:
LGTM, merge at will!

PR #3, opened 2016-07-10
Adds LBP (Local Binary Patterns)

mronian:
- [x] 3x3 LBP
- [x] Circular LBP
- [x] Uniform LBP
- [x] Modified LBP
- [x] Direction Coded LBP
- [x] Multi Block LBP
- [x] Rotation Invariant LBP
- [x] Create Feature using Histogram


src/lbp.jl
```diff
@@ -0,0 +1,92 @@
+function lbp_original(bit_pattern::Array{Bool, 1})
+	sum([b * 2 ^ (i - 1) for (i, b) in enumerate(bit_pattern)])
+end
+
+UNIFORM_LBP_TABLE = Dict{Array{Bool, 1}, Int}()
```
timholy: If you can, use `const` on these. Will make performance much better (type inference will work for `const` globals but not for non-const globals).

src/lbp.jl
```diff
@@ -0,0 +1,92 @@
...
+UNIFORM_PATTERN_COUNT = 0
+NON_UNIFORM_PATTERN = Array{Bool, 1}()
+
+function init_uniform_lbp_params(points::Integer)
+
+	global UNIFORM_LBP_TABLE, NON_UNIFORM_PATTERN, UNIFORM_PATTERN_COUNT
+	
+	UNIFORM_PATTERN_COUNT = 0
+	UNIFORM_LBP_TABLE = Dict{Array{Bool, 1}, Int}()
```
timholy: This will break if you use `const`. Can you use `empty!(UNIFORM_LBP_TABLE)`?

src/lbp.jl
```diff
@@ -0,0 +1,92 @@
+function lbp_original(bit_pattern::Array{Bool, 1})
+	sum([b * 2 ^ (i - 1) for (i, b) in enumerate(bit_pattern)])
+end
+
+UNIFORM_LBP_TABLE = Dict{Array{Bool, 1}, Int}()
+UNIFORM_PATTERN_COUNT = 0
```
timholy: If this is going to be incremented, you can't make it `const`. For performance you might want to annotate each call site with `::Int`. Alternatively, consider making this a `Ref(0)`.

src/lbp.jl
```diff
@@ -0,0 +1,92 @@
...
+	R = CartesianRange(size(img))
+	for I in R
+		bit_pattern = [img[I] < bilinear_interpolation(img, I[1] + o[1], I[2] + o[2]) ? false : true for o in offsets]
+		lbp_image[I] = method(bit_pattern)
+	end
+	lbp_image
+
+end
+
+function original_offsets()
```
timholy: This could also be a `const` (not that it matters, really).

mronian:
Tests fail due to timholy/Images.jl#532.

@timholy This is ready for review. Is there a workaround for the global variables?

Annotating each call site for the `UNIFORM_PATTERN_COUNT` variable with `::Int` raises a deprecated warning. I could change the API to pass the global variables back and forth but will that be inefficient?

mronian:
Tests pass now.

src/core.jl
```diff
@@ -0,0 +1,26 @@
+abstract Detector
+
+type Keypoint
```
timholy: I trust you'll use a single `Keypoint` throughout the package.

src/lbp.jl
```diff
@@ -0,0 +1,146 @@
+function lbp_original(bit_pattern::Array{Bool, 1})
+	sum([b * 2 ^ (length(bit_pattern) - i) for (i, b) in enumerate(bit_pattern)])
```
timholy: Using `1<<(length(bit_pattern)-i)` might be a little faster.

src/lbp.jl
```diff
@@ -0,0 +1,146 @@
...
+	NON_UNIFORM_PATTERN = zeros(Bool, points)
+	NON_UNIFORM_PATTERN[1:2:points] = true
+	UNIFORM_LBP_TABLE[NON_UNIFORM_PATTERN] = points * (points - 1) + 2
+end
+
+function lbp_uniform(bit_pattern::Array{Bool, 1})
+	global UNIFORM_LBP_TABLE, NON_UNIFORM_PATTERN, UNIFORM_PATTERN_COUNT
+	variations = sum([bit_pattern[i] != bit_pattern[i + 1] for i in 1:length(bit_pattern) - 1])
+	if variations <= 2
+		try
```
timholy: better to use `haskey`; `try/catch` as means of control flow is quite slow.

src/lbp.jl
```diff
@@ -0,0 +1,146 @@
...
+	UNIFORM_PATTERN_COUNT::Int = 0
+	empty!(UNIFORM_LBP_TABLE)
+	NON_UNIFORM_PATTERN = zeros(Bool, points)
+	NON_UNIFORM_PATTERN[1:2:points] = true
+	UNIFORM_LBP_TABLE[NON_UNIFORM_PATTERN] = points * (points - 1) + 2
+end
+
+function lbp_uniform(bit_pattern::Array{Bool, 1})
+	global UNIFORM_LBP_TABLE, NON_UNIFORM_PATTERN, UNIFORM_PATTERN_COUNT
+	variations = sum([bit_pattern[i] != bit_pattern[i + 1] for i in 1:length(bit_pattern) - 1])
```
timholy: This will allocate an array on each call. Assuming this is a performance-critical function, much better to just write this out as a loop. Or if you want to support julia-0.5 only, use the new generator syntax, which is just exactly what you wrote but without the `[]` brackets.

src/lbp.jl
```diff
@@ -0,0 +1,146 @@
...
+    end
+    mini
+end
+
+function _lbp{T<:Gray}(img::AbstractArray{T, 2}, points::Integer, offsets::Array, method::Function = lbp_original)
+	init_uniform_lbp_params(points)
+	lbp_image = zeros(UInt, size(img))
+	R = CartesianRange(size(img))
+	for I in R
+		bit_pattern = [img[I] < bilinear_interpolation(img, I[1] + o[1], I[2] + o[2]) ? false : true for o in offsets]
```
timholy: You'd probably be much better off pre-allocating the `bit_pattern` array and then storing the results in it (i.e., reusing the same array). One issue, though, is that when you use it as a key for `UNIFORM_LBP_TABLE` you'll have to make a copy (see above).

src/lbp.jl
```diff
@@ -0,0 +1,146 @@
...
+
+function lbp_uniform(bit_pattern::Array{Bool, 1})
+	global UNIFORM_LBP_TABLE, NON_UNIFORM_PATTERN, UNIFORM_PATTERN_COUNT
+	variations = sum([bit_pattern[i] != bit_pattern[i + 1] for i in 1:length(bit_pattern) - 1])
+	if variations <= 2
+		try 
+			return UNIFORM_LBP_TABLE[bit_pattern] 
+		catch KeyError 
+			UNIFORM_PATTERN_COUNT::Int += 1
+			UNIFORM_LBP_TABLE[bit_pattern] = UNIFORM_PATTERN_COUNT::Int
```
timholy: If you take my advice below about re-using `bit_pattern`, this will have to become `UNIFORM_LBP_TABLE[copy(bit_pattern)] = ...`.

timholy:
> Is there a workaround for the global variables?

Sure, in almost every case you should just create instances and pass them as an argument. That would even make your code multithreading safe, if julia gets to that point.

src/core.jl
mronian: Yeah. Forgot to change it here since it wasnt being used.

mronian:
I've added your suggestions (Thanks a lot ! :) ). I'd like to get rid of the global variables before merging this though. One solution that comes to mind is - 

``` jl
function _lbp{T<:Gray}(img::AbstractArray{T, 2}, points::Integer, offsets::Array, method::Function = lbp_original)
    init_uniform_lbp_params(points,  UNIFORM_LBP_TABLE, NON_UNIFORM_PATTERN, UNIFORM_PATTERN_COUNT)
    lbp_image = zeros(UInt, size(img))
    R = CartesianRange(size(img))
    bit_pattern = falses(length(offsets))
    for I in R
        for (i, o) in enumerate(offsets) bit_pattern[i] = img[I] >= bilinear_interpolation(img, I[1] + o[1], I[2] + o[2]) end
        lbp_image[I],  UNIFORM_LBP_TABLE, NON_UNIFORM_PATTERN, UNIFORM_PATTERN_COUNT = method(bit_pattern, UNIFORM_LBP_TABLE, NON_UNIFORM_PATTERN, UNIFORM_PATTERN_COUNT)
    end
    lbp_image
end

```

Here, except for the `lbp_uniform` method, the others will simply return the values as they were passed to them. Eg - 

``` jl
function lbp_original(bit_pattern::BitArray{1}, UNIFORM_LBP_TABLE, NON_UNIFORM_PATTERN, UNIFORM_PATTERN_COUNT)
    sum([b * 1 << (length(bit_pattern) - i) for (i, b) in enumerate(bit_pattern)]), UNIFORM_LBP_TABLE, NON_UNIFORM_PATTERN, UNIFORM_PATTERN_COUNT
end
```

@timholy Will this be efficient?

timholy:
I agree re getting rid of the global variables. What's strange is that I wrote out a concrete suggestion but it seems to have been lost (maybe I never clicked "Comment").

My suggestion is to define a type that holds your 3 global variables. You can basically turn `init_uniform_lbp_params` into the constructor for this type. Pass the instance around as a single argument, so you don't have such long argument lists.

mronian:
Oh. Cool, I'll do that and finish this :)

src/lbp.jl
```diff
@@ -0,0 +1,144 @@
+type UNIFORM_PARAMS
+	table::Dict{BitArray{1}, Int}
+	count::Int
+	non_uniform_pattern::BitArray{1}
+end
```
mronian: @timholy Does `BitArray{1}` need to be passed as a parameter to make it efficient or is this ok?

mronian: Is this correct? 

``` jl
type UNIFORM_PARAMS{B<:BitArray{1}}
    table::Dict{B, Int}
    count::Int
    non_uniform_pattern::B
end
```

Also, does `Dict` need to be passed as a parameter as well - 

``` jl
type UNIFORM_PARAMS{B<:BitArray{1}, D<:Dict{BitArray{1}, Int}}
    table::D
    count::Int
    non_uniform_pattern::B
end
```

timholy: Here's how you can tell:

``` jl
julia> isleaftype(BitArray)
false

julia> isleaftype(BitArray{1})
true
```

As long as it's a leaf type, it doesn't have to be a parameter. Your `Dict` should be fine too.

Any special reason for the all-caps? How about `LBPatternDict` or `PatternCache` or something like that?

mronian: Cool. Thanks!

I'll change the name.

PR #5, opened 2016-08-05
Adds Corner Orientation Calculation (Required for ORB)

mronian:


mronian:
@timholy I am merging this now since I need it for further functions but I have marked this with the label `needs-review` so that any such non reviewed code can be reviewed later. I'll fix the mac Travis failures later on (will probably get fixed once the `OutOfMemory()` bug is fixed).

PR #6, opened 2016-08-06
Adds ORB Descriptor (Oriented FAST and Rotated BRIEF)

mronian:


mronian:
res

Rotation Invariant ORB Keypoints and Descriptors!

src/orb.jl
```diff
@@ -0,0 +1,118 @@
...
+            push!(descriptor, img[k + pixel0] < img[k + pixel1])
+        end
+        push!(ret_keypoints, k)
+        push!(descriptors, descriptor)
+        @label endofloop
+    end
+    descriptors, ret_keypoints
+end
+
+const sampling_pattern = [[8, -3, 9, 5], [4, 2, 7, -12], [-11, 9, -8, 2], [7, -12, 12, -13], [2, -13, 2, 12], [1, -7, 1, 6],
```
timholy: I didn't see any typos here :smile:

More seriously, can this be generated somehow?

src/orb.jl
```diff
@@ -0,0 +1,118 @@
...
+    downsample::Float64
+    levels::Int
+    sigma::Float64
+end
+
+function ORB(; num_keypoints::Int = 500, n_fast::Int = 12, threshold::Float64 = 0.25, harris_factor::Float64 = 0.04, downsample::Real = 1.3, levels::Int = 8, sigma::Float64 = 1.2)
+    ORB(num_keypoints, n_fast, threshold, harris_factor, downsample, levels, sigma)
+end
+
+function gaussian_pyramid{T}(img::AbstractArray{T, 2}, n_scales::Int, downsample::Real, sigma::Real)
```
timholy: duplicate (you have many PRs...)

timholy:
That's really exciting!

src/orb.jl
mronian: Yep. I needed this for the tests to pass :)

src/orb.jl
mronian: Actually, these patterns are learnt after training the detector on large datasets. I have taken these values from opencv directly. Same will be the case for FREAK descriptors :/

mronian: Should I move these to a text file or something?

timholy: Up to you. But that's not a crazy idea.

mronian: Will reading them from a text file each time a descriptor needs to be calculated take up too much time?

timholy: Yes.

I was just thinking that you meant having a `consts.jl` file that you `include`, so they don't clutter up the code; those consts would get precompiled in.

mronian: Oh :sweat_smile: !

Yes that seems better. :)

mronian:
@timholy I wanted to clarify how the MIT & BSD Licenses work. Since I have taken the sampling_pairs from opencv, should I include the BSD License?

PR #9, opened 2016-08-14
Adds FREAK Keypoints and Descriptors

mronian:


mronian:
@timholy This is ready for review. Adding tests now.

mronian:
Needs blockwise hamming matcher for speedup.

src/freak.jl
```diff
@@ -0,0 +1,105 @@
+type FREAK{S, T, O, N} <: DescriptorParams
+    pattern_scale::Float64
+    pattern_table::Array{S, N}
+    smoothing_table::Array{T, N}
+    orientation_weights::Array{O}
```
timholy: Should be `Vector{O}` or `Array{O,1}`. (Not a leaftype otherwise.)

src/freak.jl
```diff
@@ -0,0 +1,105 @@
...
+        push!(orientation_weights, OrientationPair([round(Int, dy * 4096 / norm), round(Int, dx * 4096 / norm)]))
+    end
+    FREAK(pattern_scale, pattern_table, smoothing_table, orientation_weights)
+end
+
+function _freak_mean_intensity{T<:Gray}(int_img::AbstractArray{T, 2}, keypoint::Keypoint, offset::SamplePair, sigma::Float64)
+    y = k[1] + offset[1]
+    x = k[2] + offset[2]
+    if sigma < 0.5
+        tl = bilinear_interpolation(int_img, y - sigma, x - sigma)
```
timholy: In the new ImagesFiltering, I think we should make `IntegralImage` a type and try to give it a nice interface. So that you could write this entire function something like this:

``` jl
if σ < 0.5
    return interpolate(int_img, I±σ)
end
int_img[I±σ]
```

where `I` is a `CartesianIndex`.

This is pretty cute:

``` jl
julia> ±(x, y) = (x-y, x+y)
± (generic function with 1 method)

julia> 5±3
(2,8)
```

src/freak.jl
```diff
@@ -0,0 +1,105 @@
...
+        direction_sum_y += orientation_weights[i] * intensity_diff / 4096
+        direction_sum_x += orientation_weights[i] * intensity_diff / 4096
+    end
+    angle = atan2(direction_sum_y, direction_sum_x)
+    scaled_angle = round(Int, (angle + pi) * freak_orientation_steps / (2 * pi))
+    scaled_angle
+end
+
+function _freak_tables(pattern_scale::Float64)
+    pattern_table = Array{SamplePair}[]
```
timholy: Remember the dimensional parameter, or it's not a leaftype.

timholy: In case you're confused, again remember that function signatures are totally different from creation in terms of whether it matters. (Functions get specialized on the types that actually get passed in, so are automatically using leaf types if that's what you pass in. Type declarations and created containers do exactly what you tell them.)

src/freak.jl
```diff
@@ -0,0 +1,105 @@
...
+        end
+        push!(pattern_table, pattern)
+        push!(smoothing_table, sigmas)
+    end
+    pattern_table, smoothing_table
+end
+
+function create_descriptor{T<:Gray}(img::AbstractArray{T, 2}, keypoints::Keypoints, params::FREAK)
+    int_img = integral_image(img)
+    descriptors = BitArray[]
```
timholy: missing parameter

src/freak.jl
mronian: Seems great!

mronian:
@timholy One thing I realized while doing FREAK and BRISK is that they work best with AGAST corners which we do not have yet. The current tests account for rotation and translation invariance but for scale invariance we will need AGAST corners. I'll do those once GSoC is done. 

P.S. I've incorporated your suggestions :)

src/freak.jl
```diff
@@ -0,0 +1,105 @@
+type FREAK{S, T, O} <: DescriptorParams
+    pattern_scale::Float64
+    pattern_table::Array{Array{S, 1}, 1}
```
mronian, line 3: Is this correct?

timholy, line 3: Yes

timholy:
LGTM

src/freak.jl
```diff
@@ -0,0 +1,105 @@
...
+
+function _freak_mean_intensity{T<:Gray}(int_img::AbstractArray{T, 2}, keypoint::Keypoint, offset::SamplePair, sigma::Float64)
+    y = keypoint[1] + offset[1]
+    x = keypoint[2] + offset[2]
+    if sigma < 0.5
+        tl = bilinear_interpolation(int_img, y - sigma, x - sigma)
+        tr = bilinear_interpolation(int_img, y - sigma, x + sigma)
+        bl = bilinear_interpolation(int_img, y + sigma, x)
+        br = bilinear_interpolation(int_img, y + sigma, x + sigma)
+        return (br + tl - tr - bl) / (4 * (sigma ^ 2))
```
mronian: @timholy This wasnt covered by the test cases since default `pattern_scale` is 22 which scales up the `sigma`. After checking coveralls I attempted to cover this condition too. Turns out this is not correct as bilinear interpolation of the integral image may not give correct values to calculate a `diff`. What would be a good alternative for this? Should I just take an average of the 8x8 neighbourhood?

test/freak.jl
```diff
@@ -0,0 +1,77 @@
...
+	    img_temp_2 = _warp(img_array_1, 5 * pi / 6)
+	    img_array_2 = _warp(img_temp_2, 50, 40)
+
+        keypoints_1 = Keypoints(fastcorners(img_array_1, 12, 0.35))
+        keypoints_2 = Keypoints(fastcorners(img_array_2, 12, 0.35))
+        freak_params = FREAK()
+
+        desc_1, ret_keypoints_1 = create_descriptor(img_array_1, keypoints_1, freak_params)
+        desc_2, ret_keypoints_2 = create_descriptor(img_array_2, keypoints_2, freak_params)
+        matches = match_keypoints(ret_keypoints_1, ret_keypoints_2, desc_1, desc_2, 0.1)
```
mronian, line 22: Will correct this indentation issue.

src/freak.jl
timholy: I would say the best would be to write out the formula for the mean intensity over the chosen region in terms of the original (pre-integral) image. You can of course calculate those original pixel intensities from the integral image.

This is another good reason to turn your integral image into a type: there's no sense in burying these calculations inside a method specific to FREAK, instead you want to make them available to anyone who might want to use them.

If you want to give it a stab but don't quite know how to start, just ask. Some of my packages like `IndirectArrays` and `MappedArrays` give small, self-contained examples showing how to create a new AbstractArray type.

mronian: I agree. This is required in BRISK for intensity sampling too and it makes much more sense to have it in `ImageFiltering`. I'll change this to a mean intensity for now and open up an issue in `ImageFiltering` for the `IntegeralImage` type since I will need help in that and it would be a better place to discuss.

PR #11, opened 2016-08-17
BRISK Descriptors

mronian:


mronian:
Keypoints willl require AGAST corners. 

@timholy This is ready for review. I have added the `Feature` type. While calculating matches using `match_keypoints`, I am not sure whether to have features being passed as arguments or keypoints. Should it be `match_features`? Currently it returns an Array of (Keypoint, Keypoint) arrays. I guess returning two separate arrays would be better...

test/brisk.jl
```diff
@@ -0,0 +1,77 @@
...
+        img_temp_2 = _warp(img_array_1, pi / 4)
+        img_array_2 = _warp(img_temp_2, 10, 20)
+
+        features_1 = Features(fastcorners(img_array_1, 12, 0.2))
+        features_2 = Features(fastcorners(img_array_2, 12, 0.2))
+        brisk_params = BRISK()
+
+        desc_1, ret_features_1 = create_descriptor(img_array_1, features_1, brisk_params)
+        desc_2, ret_features_2 = create_descriptor(img_array_2, features_2, brisk_params)
+        matches = match_keypoints(Keypoints(ret_features_1), Keypoints(ret_features_2), desc_1, desc_2, 0.1)
```
mronian, line 40: This is the line I was talking about. Should we accept features instead?

src/core.jl
```diff
@@ -1,13 +1,30 @@
-abstract Detector
-abstract DescriptorParams
+abstract Params
```
mronian, line 1: Changed `DescriptorParams` to `Params` as many algorithms do the work of Detection and Description both, examples being ORB and BRISK.

src/brisk.jl
```diff
@@ -0,0 +1,100 @@
...
+    smoothing_table::Vector{T}
+    orientation_weights::Vector{O}
+    short_pairs::Vector{O}
+    long_pairs::Vector{O}
+end
+
+typealias OrientationPair Vector{Int}
+typealias SamplePair Vector{Float64}
+
+function BRISK(; threshold::Float64 = 0.25, octaves::Int = 4, pattern_scale = 1.0)
```
mronian: The `threshold` and `octaves` are for scale invariance. They will be used once AGAST is added.

src/brisk.jl
```diff
@@ -0,0 +1,100 @@
+type BRISK{S, T, O} <: Params
+    threshold::Float64
+    octaves::Int
+    pattern_scale::Float64
+    pattern_table::Vector{Vector{S}}
+    smoothing_table::Vector{T}
+    orientation_weights::Vector{O}
+    short_pairs::Vector{O}
+    long_pairs::Vector{O}
```
mronian: This looks better. I'll change it to Vectors in FREAK too.

src/brisk.jl
```diff
@@ -0,0 +1,109 @@
...
+    smoothing_table = Vector{Float64}[]
+    for ori in 0:brisk_orientation_steps - 1
+        theta = ori * 2 * pi / brisk_orientation_steps 
+        pattern = SamplePair[]
+        sigmas = Float64[]
+        for (i, n) in enumerate(brisk_num_circular_pattern)
+            for circle_number in 0:n - 1
+                angle = (circle_number * 2 * pi / n) + theta
+
+                push!(pattern, SamplePair([brisk_radii[i] * sin(angle) * pattern_scale * 0.85,
```
timholy: If these are always 2-vectors, you'll probably be better off with

``` jl
typealias SamplePair Tuple{Float64,Float64}
```

or even making an immutable. Each array has some overhead associated with it, I think ~200 bytes, and an array-of-arrays is basically an array of pointers, each of which points to another array. In contrast, a `Tuple{Float64,Float64}` (or the equivalent `immutable`) is just 16 bytes, and they are densely packed in the array by value (i.e., you don't need a pointer dereference).

timholy: (But don't just make that change, first `@time` it and `@profile` it to see whether the allocation of these arrays is a bottleneck. I haven't done this on very much of your code myself, so I don't know what the answer will be.)

src/brisk.jl
```diff
@@ -0,0 +1,109 @@
...
+    pattern_table, smoothing_table
+end
+
+function create_descriptor{T<:Gray}(img::AbstractArray{T, 2}, features::Array{Feature}, params::BRISK)
+    int_img = integral_image(img)
+    descriptors = BitArray{1}[]
+    ret_features = Feature[]
+    window_size = ceil(Int, (brisk_radii[end] + brisk_sigma[end]) * params.pattern_scale * 0.85) + 1
+    tl_lim = CartesianIndex(-window_size, -window_size)
+    br_lim = CartesianIndex(window_size, window_size)
```
timholy: You can just define one of these and use `keypoint + lim` and `keypoint - lim`.

src/brisk.jl
```diff
@@ -0,0 +1,109 @@
...
+
+function _brisk_orientation{T<:Gray}(int_img::AbstractArray{T, 2}, keypoint::Keypoint, pattern::Array{SamplePair}, 
+                                        orientation_weights::Array{OrientationPair}, sigmas::Array{Float64}, long_pairs::Array{OrientationPair})
+    direction_sum_y = 0.0
+    direction_sum_x = 0.0
+    for (i, o) in enumerate(long_pairs)
+        offset_1 = pattern[o[1]]
+        offset_2 = pattern[o[2]]
+        intensity_diff = ImageFeatures._freak_mean_intensity(int_img, keypoint, offset_1, sigmas[o[1]]) - ImageFeatures._freak_mean_intensity(int_img, keypoint, offset_2, sigmas[o[2]])
+        direction_sum_y += orientation_weights[i][1] * intensity_diff / 4096
```
timholy: what's the 4096 about?

test/brisk.jl
timholy, line 40: You don't have to choose; just define an alternate form of `match_keypoints` that takes a `Feature` input and extracts the `Keypoint`, then calls the version of `match_keypoints` accepting a `Keypoint` input.

src/brisk.jl
```diff
@@ -0,0 +1,109 @@
...
+                push!(short_pairs, OrientationPair([j, i]))
+            end
+        end
+    end
+    BRISK(threshold, octaves, pattern_scale, pattern_table, smoothing_table, orientation_weights, short_pairs, long_pairs)
+end
+
+function _brisk_orientation{T<:Gray}(int_img::AbstractArray{T, 2}, keypoint::Keypoint, pattern::Array{SamplePair}, 
+                                        orientation_weights::Array{OrientationPair}, sigmas::Array{Float64}, long_pairs::Array{OrientationPair})
+    direction_sum_y = 0.0
```
timholy: Any reason the signature couldn't be
`_brisk_orientation{T<:Gray}(int_img::AbstractArray{T,2}, keypoint, pattern, orientation_weights, sigmas, long_pairs)`?
In other words, do all the types add something?

src/brisk.jl
mronian: I was trying to reduce storage in the `orientation_weights` vector by using Int values instead of Float. Now that I think of it, Int by default takes the same storage as Float right (64bit)? I will have to explicitly specify Int16?

test/brisk.jl
mronian, line 40: I had thought of that but then should the API be `match_keypoints` (could be misleading)....I am thinking of changing it to `find_matches` which will take a `matcher_type` argument which in this case would be `hamming_distance`. Later we could expand to Flann etc.

src/brisk.jl
mronian: No. I guess types wont help in non-exported functions as they are not meant to be used directly anyways. I had added the types so that in case someone tries to give wrong arguments, the error message will be useful.

src/brisk.jl
timholy: Right. Another good option is `Float16`, a number type which seems to make an awful lot of sense in image applications but which we do not use heavily now. Until very recently, computations with `Float16` promoted to `Float32`, and some (like multiplication by an `Integer`) still do, but overall it's become a fairly satisfying type to use for storage and computation.

src/brisk.jl
timholy: There are cases where the typing does generate helpful error messages, so you can leave them if you prefer. I'm wondering if some day someone might want to choose a subset of keypoints with a `SubArray`, though.

mronian: Would it be better if we change `Keypoints` to be an `AbstractVector`?

src/brisk.jl
mronian: Since I am using the weights and sampling pairs in many mathematical operations, I guess using Tuple wont be possible as math operations are not allowed on Tuples :(

mronian: Made the modifications just to check the benchmark scores - 

For the `BRISK()` call - 

```
# With Vector 

  samples:          238
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  16.66 mb
  allocs estimate:  350205
  minimum time:     14.51 ms (0.00% GC)
  median time:      22.13 ms (30.47% GC)
  mean time:        21.08 ms (26.61% GC)
  maximum time:     28.05 ms (32.27% GC)

# With Tuple

  samples:          694
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  1.15 mb
  allocs estimate:  12339
  minimum time:     6.54 ms (0.00% GC)
  median time:      7.02 ms (0.00% GC)
  mean time:        7.20 ms (2.53% GC)
  maximum time:     12.12 ms (28.43% GC)
```

For the `create_descriptor` call - 

```
# With Vector

  samples:          5
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  299.36 mb
  allocs estimate:  18317758
  minimum time:     998.75 ms (3.71% GC)
  median time:      1.01 s (3.64% GC)
  mean time:        1.02 s (3.58% GC)
  maximum time:     1.04 s (3.75% GC)

# With Tuple

  samples:          6
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  274.89 mb
  allocs estimate:  17556044
  minimum time:     996.81 ms (2.77% GC)
  median time:      999.79 ms (2.87% GC)
  mean time:        999.19 ms (2.88% GC)
  maximum time:     1.00 s (3.00% GC)
```

I guess there is quite an improvement in the memory storage in the `BRISK` call (I wasn't expecting much in `create_descriptor` anyways as it wasn't affected much by the conversion to `Tuple`). I am changing the `Vector` to `Tuple` :)

src/brisk.jl
```diff
@@ -0,0 +1,106 @@
...
+"""
+type BRISK <: Params
+    threshold::Float64
+    octaves::Int
+    pattern_scale::Float64
+    pattern_table::Vector{Vector{SamplePair}}
+    smoothing_table::Vector{Vector{Float16}}
+    orientation_weights::Vector{OrientationWeights}
+    short_pairs::Vector{OrientationPair}
+    long_pairs::Vector{OrientationPair}
```
mronian, line 18: Since these variables are internal, I've replaced the parametric `BRISK{}` as this looks better. All of them are leaftypes so I guess it shouldnt matter? At some point, we should have redefine the `print` function for these types so that the internal variables do not get printed while displaying the result of calling the constructor.

src/brisk.jl
timholy: Another option is either FixedSizeArrays or StaticArrays (I'm switching to the latter). These are effectively "tuples with math operations supported." Images-next is using StaticArrays already in ImagesFiltering.

src/brisk.jl
timholy: Right, as long as you're using leaftypes, it doesn't matter whether you have parameters or not.

PR #19, opened 2016-08-24
BRISK tutorial

mronian:


mronian:
@timholy The build on julia-0.5 takes an awful lot of time on the test cases for some reason and hence it always fails. I put in some print statements to get it to complete and it took [almost an hour!](https://travis-ci.org/JuliaImages/ImageFeatures.jl/builds/154729399) On the nightly build though it takes around 20 minutes. Any idea what could be the problem? I have run the tests on my local machine with both julia-0.6 and 0.5-rc3 and the tests run pretty fast.

timholy:
Hmm. Looks like all that time was spent building Rmath, which comes in only because of `StatsBase.histrange`. I think this is the final nail in the coffin: while generally I like the idea of reusing code, I'll backtrack on my own earlier advice and propose that we switch to our own internal `histrange` function.

timholy:
Just to check that the images of the sampling patterns in your ORB and FREAK tutorials are unencumbered, right? (Do you need to give image credits?)

mronian:
I'm not sure it is because of building Rmath.....It was only when it reached the ORB test it got stuck (Line 532 here https://travis-ci.org/JuliaImages/ImageFeatures.jl/jobs/154729400#L532). Each of the print statements (Succes1, Success2 etc) took like 5-10 minutes each. On the nightly build the entire test suite finishes by 20 minutes.

mronian:
Well, I have taken the images from TestImages...I think the `Lena` image had some issues so I'll change that. The `lighthouse` image was in `TestImages` earlier so I am not sure if it had copyright issues or not.

timholy:
Re the images: I meant the images of the sampling patterns, i.e., the plots of the circles at various positions.

You're right, I must have misinterpreted the meaning of that time marker. A much more likely culprit is the fact that package tests, [when you ask for coverage](https://github.com/JuliaImages/ImageFeatures.jl/blob/d11f4a0748ba1a7327472e49440806e5419eba24/.travis.yml#L17), are run with `--inline=no` (see `Pkg.test` code [here](https://github.com/JuliaLang/julia/blob/f58aaef9b7aa202e9d59e59798ce7f5c32706a1e/base/pkg/entry.jl#L706)). This is because coverage is woefully inaccurate if inlining is on. If you run the tests locally with `--inline=no`, then it indeed seems pretty slow.

There's work going on in base julia (https://github.com/JuliaLang/julia/pull/18196) to give accurate coverage statistics even with inlining on. So in 0.6 this shouldn't be an issue. In the meantime, I think you'll have to choose between using much smaller images in the tests (fewer data points = faster execution) or turning off coverage.

Issue #24, opened 2017-03-04
Error in Test Case


mronian:
As pointed out in https://github.com/JuliaImages/Images.jl/pull/601#issuecomment-284150087.

mronian:
Are there other test failures too?

timholy:
Sometimes you do have to change the test values when algorithms update. In https://github.com/JuliaImages/Images.jl/pull/596 I found a test that essentially "froze in" the exact behavior of the algorithm we'd had before, rather than more broadly testing "sensible behavior." The tricky part is that sometimes it's pretty hard to define "sensible."

@annimesh2809, I'd say begin by commenting out failing tests and then we can discuss.

annimesh2809:
@mronian Sorry I did not see your comment so didn't reply sooner.

I have fixed some warnings and I have printed the value returned by the algorithm and the expected value of a test case.

![screenshot from 2017-03-04 19 44 37](https://cloud.githubusercontent.com/assets/15063205/23579431/01926852-0113-11e7-944a-5888042ba3af.png)

The failing test case is at line 55 of /test/brief.jl.
What to do with this test case?

mronian:
Its a very simple test case which is already covered by the ones after it in the `Descriptor Calculation` part. You can comment it out for now. I'll submit a modification.

annimesh2809:
Another failing test on line 253 and 259 of file /test/lbp.jl. Both the results of "Descriptor" are different from expected value. Here is a screenshot of the test failure and the returned vs expected value:

![img](https://cloud.githubusercontent.com/assets/15063205/23579970/bb41026c-011e-11e7-9ee5-a6abef05a4b4.png)

annimesh2809:
Initial tests in test/corner.jl call the `corner_orientations(img::AbstractArray)` function from line 49, ImageFeatures/src/corner.jl. Here the function `Kernel.gaussian` (previously `gaussian2d`) is called from line 134, ImageFiltering/src/kernel.jl which calls `KernelFactors.gaussian` from line 269, ImageFiltering/kernelfactors.jl and returns an object of type `OffsetArray` by a call to `centered` from line 14, ImageFiltering/src/utils.jl. The problem arises due to the fact that the function
`corner_orientations(img::AbstractArray{T, 2}, corners::Keypoints, kernel::Array{K, 2})`
[...]

PR #25, opened 2017-03-05
Patching ImageFeatures.jl

annimesh2809:
ImageFeatures.jl had some issues with the new Images.jl as discussed in #24 

Contains patches for warnings and errors. Now package test `Pkg.test("ImageFeatures")` runs without warning or errors on Julia v0.5 with the new images. But the coverage is reduced due to 2 failing tests present in "Descriptor" in /test/lbp.jl which have been commented out.

Coverage report with test cases of orb.jl -> (test cases took time>30 mins) and some test cases of lbp.jl (as mentioned above) commented out:
img2


src/brief.jl
```diff
@@ -148,7 +148,11 @@ function centered(size::Int, window::Int, seed::Int)
 end
 
 function create_descriptor{T<:Gray}(img::AbstractArray{T, 2}, keypoints::Keypoints, params::BRIEF)
-    img_smoothed = imfilter_gaussian(img, [params.sigma, params.sigma])
+    
+    _eltype{T,C<:Colorant}(::Type{T}, ::Type{C}) = base_colorant_type(C){T}
+    _eltype{T,R<:Real}(::Type{T}, ::Type{R}) = T
+    factkernel = KernelFactors.IIRGaussian([params.sigma, params.sigma])
+    img_smoothed = imfilter(_eltype(Float64, eltype(img)), img, factkernel, NA())
```
timholy: Since [`params.sigma` is already a `Float64`](https://github.com/JuliaImages/ImageFeatures.jl/blob/d11f4a0748ba1a7327472e49440806e5419eba24/src/brief.jl#L18), doesn't it choose that eltype correctly already? If not, I should fix something in ImageFiltering.

src/corner.jl
```diff
@@ -19,7 +19,7 @@ function corner_orientations{T<:Gray, K<:Real}(img::AbstractArray{T, 2}, corners
 	pre_x = ceil(Int, (w - 1) / 2)
 	post_y = floor(Int, (h - 1) / 2)
 	post_x = floor(Int, (w - 1) / 2)
-	img_padded = padarray(img, [pre_y, pre_x], [post_y, post_x], "value", 0)
+    img_padded = padarray(img, Fill(0, (pre_y, pre_x), (post_y, post_x))).parent
```
timholy: Line spacing (minor). More importantly, better use `parent(padarray(...))` than accessing the `parent` field: otherwise, this will fail if `padarray` ever changes to return a type that doesn't have a `parent` field. (As a function, `parent` has a fallback definition to just return the array itself.)

There are other places like this below; they should be fixed too (but I won't comment there).

test/brief.jl
```diff
@@ -12,7 +12,7 @@ facts("BRIEF") do
     context("Sampling Patterns") do
         # Check Bounds
         for si in [32, 128, 256], wi in [5, 15, 25, 49], se in [123, 546, 178]
-            s1, s2 = centered(si, wi, se)
+            s1, s2 = ImageFeatures.centered(si, wi, se)
```
timholy: Oh wow, I hadn't remembered that ImageFeatures exports a `centered` function (ImageFiltering now exports the same function).

How to solve this? @mronian, would you be against having us group all the sampling patterns into a `Brief` module? So this would become `Brief.centered(si, wi, se)` and ImageFeatures would export `Brief` but not `Brief.centered`. Sort of how ImageFiltering exports `Kernel` and `KernelFactors` modules but none of the specific kernels themselves.

src/glcm.jl
```diff
@@ -12,12 +12,12 @@ function glcm{T<:Real, A<:Real}(img::AbstractArray{T, 2}, distances::Array{Int,
 end
 
 function glcm{T<:Colorant}(img::AbstractArray{T, 2}, distance::Integer, angle::Real, mat_size::Integer = 16)
-    img_rescaled = map(i -> max(1, Int(ceil((convert(Gray{U8}, i).val.i) * mat_size / 256))), img)
+    img_rescaled = map(i -> max(1, Int(ceil((convert(Gray{FixedPointNumbers.Normed{UInt8, 8}}, i).val.i) * mat_size / 256))), img)
```
timholy: `N0f8` is the short-form version of `FixedPointNumbers.Normed{UInt8, 8}` (see [docs](http://juliaimages.github.io/latest/quickstart.html#Colors,-the-0-to-1-intensity-scale,-and-views-1) and [more docs](http://juliaimages.github.io/latest/arrays_colors.html)). And rather than `convert(Gray{N0f8}, i).val.i` it should be `reinterpret(gray(convert(Gray{N0f8}, i)))`.

test/brief.jl
```diff
@@ -27,22 +27,22 @@ facts("BRIEF") do
     end
 
     context("Descriptor Calculation") do
-        img_1 = Gray{Float64}[ 0 0 0 0 0 0 
-                  0 0 1 1 0 0 
-                  0 0 1 1 0 0 
-                  0 0 1 1 0 0 
-                  0 0 1 1 0 0 
-                  0 0 0 0 0 0 ]
+        #=img_1 = Gray{Float64}[  0 0 0 0 0 0
```
timholy: The fundamental problem with this test is that `imcorner` seems to be changing what it returns. The old test seems to have relied on the fact that the result of `harris` on this input image had a tiny numerical difference between location `5,3` and `5,4` (due to roundoff error); with the new Images, the two are exactly the same, and hence `findlocalmaxima` doesn't return it.

I think the best thing would be to widen the array and make sure there are 3 columns of 1s. That would make the test much more robust. OK, @mronian?

test/lbp.jl
```diff
@@ -245,18 +245,22 @@ facts("Local Binary Patterns") do
 		@fact multi_block_lbp(img, 1, 1, 4, 4) --> 18
 	end
 
-	context("Descriptor") do
+	#=context("Descriptor") do
```
timholy: I'm not sure what to do about this one. @mronian?

test/orb.jl
```diff
@@ -33,7 +33,7 @@ facts("ORB") do
         desc_2, ret_keypoints_2 = create_descriptor(img_array_2, orb_params)
         matches = match_keypoints(ret_keypoints_1, ret_keypoints_2, desc_1, desc_2, 0.2)
         reverse_keypoints_1 = [_reverserotate(m[1], pi / 4, (256, 384)) for m in matches]
-        @fact all(isapprox(rk[1], m[2][1], atol = 3) && isapprox(rk[2], m[2][2], atol = 3) for (rk, m) in zip(reverse_keypoints_1, matches)) --> true
```
timholy, line 36: Seems like this has one failure out of 72. Should we relax the test by requiring that 95% line up?

timholy (commented):
Aside from the commented-out tests and a few issues I noted, this looks ready to go. Nice work! @mronian, I've pinged you at a couple of locations below.

test/brief.jl
mronian: I'm cool with that. Will it be an issue though if we just change the name of the function to something else?

test/brief.jl
mronian: Yeah, that should work. @annimesh2809 you could see the tests which follow this one for more examples. To add variety, we can add some other shape like a triangle since the other tests cover this one.

test/orb.jl
mronian: @annimesh2809, @timholy how about adding a function like `grade_matches` or something which takes in two sets of keypoints and returns a score for the quality of matches? It will be useful for object tracking also.

test/brief.jl
timholy: That would be fine, too. We could alternatively change the one in ImageFiltering, but to me (and I'm definitely biased) it seems that one is a little more general.

test/lbp.jl
mronian: To make them robust, it can be replaced by a binary image whose descriptor is manually calculated. 

@annimesh2809 If you would like to do this, let me know. If not, I'll do this one 😄

src/brief.jl
annimesh2809: Yes, the eltype is correctly chosen so we don't need the `_eltype` function to do that for us. I will modify the code as required.

test/lbp.jl
annimesh2809: @mronian I'll try to do it. By "manually calculated", do you mean that I have to calculate the LBP mask by hand for the whole image?

.travis.yml
```diff
@@ -14,7 +14,7 @@ script:
     - julia -e 'Pkg.add("Images")'  # Needed for Docs 
     - julia -e 'Pkg.clone("https://github.com/JuliaImages/ImageDraw.jl")' # Needed for Docs 
     - julia -e 'Pkg.clone(pwd()); Pkg.build("ImageFeatures")'
-    - julia -e 'Pkg.test("ImageFeatures", coverage=true)'
+    - julia -e 'Pkg.test("ImageFeatures",coverage=true)'
```
mronian, line 17: ?

src/core.jl
```diff
@@ -104,4 +104,27 @@ function match_keypoints(keypoints_1::Keypoints, keypoints_2::Keypoints, desc_1,
...
+"""
+```
+grade = grade_matches(keypoints_1, keypoints_2, difference_method)
+```
+
+Returns a measure of similarity between keypoints in `keypoints_1` and `keypoints_2` where `difference_method`
+is the method used for computing difference between individual pair of keypoints.
+"""
+
+function grade_matches(K1::Keypoints, K2::Keypoints, diff::Function = (i,j)->(abs(i[1] - j[1]) + (abs(i[2] - j[2]))))
```
mronian: It would be better if the default distance is euclidean distance. @timholy Should we use [Distances.jl](https://github.com/JuliaStats/Distances.jl) here?

mronian: You need to check if the lengths of K1 and K2 are equal.

.travis.yml
annimesh2809: Because the tests for orb.jl were taking too much time on travis, I used the `travis_wait` command. But the space present here caused problems with travis build, so I had to remove it.

src/core.jl
```diff
@@ -104,4 +104,27 @@ function match_keypoints(keypoints_1::Keypoints, keypoints_2::Keypoints, desc_1,
...
+    count = 0
+    for (i,j) in zip(K1, K2)
+        sum += diff(i,j)
+        count += 1
+    end
+    if count != 0
+        sum / count
+    else
+        0
+    end
```
mronian: This can be replaced by `mean(map((kp1, kp2) -> diff(kp1, kp2), K1, K2))`. Just to maintain consistency in the package could you rename K1, K2 as keypoints_1, keypoints_2?

src/core.jl
mronian:
```
@assert length(k1)==length(k2) "Keypoint lists are of different lengths."
@assert length(k1)==0 "Keypoint list is of size zero."
```
@timholy will `@assert` slow this down too much? Assuming this function might be used with a continuous video stream for tracking keypoints, we need it to be fast.

test/brief.jl
mronian: Of course. @annimesh2809 how about we change this to `centre_sample`?

test/brief.jl
```diff
@@ -27,31 +27,48 @@ facts("BRIEF") do
...
         brief_params = BRIEF(size = 8, window = 3, seed = 123)
 
         desc_1, ret_keypoints_1 = create_descriptor(img_1, keypoints_1, brief_params)
         desc_2, ret_keypoints_2 = create_descriptor(img_2, keypoints_2, brief_params)
         matches = match_keypoints(ret_keypoints_1, ret_keypoints_2, desc_1, desc_2)
-        expected_matches = [[CartesianIndex((2,3)),CartesianIndex((2,3))],
-                            [CartesianIndex((5,3)),CartesianIndex((7,3))]]
+        expected_matches = [[CartesianIndex((5,4)),CartesianIndex((3,3))],
+                            [CartesianIndex((16,4)),CartesianIndex((8,3))],
+                            [CartesianIndex((16,15)),CartesianIndex((8,8))]]
```
mronian, line 71: Nice 😄

test/lbp.jl
```diff
@@ -14,7 +14,7 @@ facts("Local Binary Patterns") do
 			 0x1a  0xe0  0x52  0x5a  0x6a  0x03  0xe8  0xcb  0x95  0xfc
 			 ]
 
-	img_gray = map(i -> Images.Gray(reinterpret(U8, i)), img)
+	img_gray = map(i -> Images.Gray(reinterpret(FixedPointNumbers.Normed{UInt8, 8}, i)), img)
```
mronian: This could use the shorthand notation used above.

src/core.jl
annimesh2809: @mronian Even though lengths of K1 and K2 are equal they might be at very different angles, so shouldn't we check for closeness between keypoints rather than only comparing their length. The [previous tests](https://github.com/JuliaImages/ImageFeatures.jl/blob/master/test/orb.jl#L36) also compared for closeness by comparing both `rk[1]` to `m[2][1]` and `rk[2]` to `m[2][2]` with the `isapprox` function.

test/lbp.jl
mronian: Yeah. You can take a simple binary image (1s and 0s) say of size 10x10. For a block size of 5x5, we will have 4 histograms which will be created from the LBP images of each 5x5 image. Appending this 4 histograms will give you the descriptor.

test/lbp.jl
mronian: Same goes for other places below.

src/core.jl
mronian: A nice addition to this would be a version of this function which takes in a `limit` parameter and returns the percentage of keypoints which have distance under this limit.

.travis.yml
timholy: Are they faster if we turn off coverage?

src/core.jl
timholy: No, such checks should be fine.

src/core.jl
```diff
@@ -104,4 +104,27 @@ function match_keypoints(keypoints_1::Keypoints, keypoints_2::Keypoints, desc_1,
...
+    sum = 0.0
+    count = 0
+    for (i,j) in zip(K1, K2)
+        sum += diff(i,j)
+        count += 1
+    end
+    if count != 0
+        sum / count
+    else
+        0
```
timholy: Not type stable. Make it `0.0`.

src/core.jl
timholy: I'm fine with using Distances, although `sum(abs2.(i-j))` is pretty easy.

src/core.jl
timholy: But of course the suggestion to use `mean(f, ...)` is even better.

src/lbp.jl
```diff
@@ -134,7 +134,7 @@ function _create_descriptor{T<:FixedPointNumbers.UFixed}(img::AbstractArray{Gray
     descriptor
 end
 
-function create_descriptor{T<:FixedPointNumbers.UFixed}(img::AbstractArray{Gray{T}, 2}, yblocks::Integer = 4, xblocks = 4; lbp_type::Function = lbp, args...)
+function create_descriptor{T<:FixedPointNumbers.Normed}(img::AbstractArray{Gray{T}, 2}, yblocks::Integer = 4, xblocks = 4; lbp_type::Function = lbp, args...)
```
timholy: Not necessary to change, but one can get rid of the `FixedPointNumbers` scoping and just use `Normed` directly. (As long as you've said `using FixedPointNumbers` rather than `import FixedPointNumbers`.)

.travis.yml
annimesh2809: Yes, they run quite fast with coverage off, but tend to take >45 mins (which is maximum time allowed by travis per job) if the coverage is switched on.

src/core.jl
```diff
@@ -104,4 +104,25 @@ function match_keypoints(keypoints_1::Keypoints, keypoints_2::Keypoints, desc_1,
...
+`difference_method(keypoint_1,keypoint_2)` less than `limit`.
+"""
+
+function grade_matches(keypoints_1::Keypoints, keypoints_2::Keypoints, diff::Function = (i,j) -> (sqrt( (i[1]-j[1])^2 + (i[2]-j[2])^2 )))
+    @assert length(keypoints_1)==length(keypoints_2) "Keypoint lists are of different lengths."
+    @assert length(keypoints_1)!=0 "Keypoint list is of size zero."
+    mean(map((keypoint_1,keypoint_2)->diff(keypoint_1,keypoint_2), keypoints_1, keypoints_2))
+end
+
+grade_matches(keypoints_1::Keypoints, keypoints_2::Keypoints, limit::Real, diff::Function = (i,j)->(sqrt( (i[1]-j[1])^2 + (i[2]-j[2])^2 ))) =
```
timholy: Do we even need this function? If we want it, it should return a `Bool` not an `Int`.

timholy (approved):
Once you two decide about that last test, I'm fine with this (see very small comment). @mronian should do the honors when it comes time to merge.

annimesh2809:
@mronian I don't understand this [line](https://github.com/JuliaImages/ImageFeatures.jl/blob/master/src/lbp.jl#L128). Shouldn't image partitioning into blocks have been done like this: `img[(j-1)*blockh+1 : j*blockh, (i-1)*blockw+1 : i*blockw]` ?

test/orb.jl
```diff
@@ -33,7 +33,7 @@ facts("ORB") do
...
+        @fact (grade_matches(reverse_keypoints_1, [match[2] for match in matches], (i,j)->(abs(i[1]-j[1]) + abs(i[2]-j[2]))) < 1.5 ) --> true
```
mronian: Use the function with `limit` here. Do 95% keypoints lie under a 10 pixel radius?

mronian:
@annimesh2809 I agree. You will have to change that 😅

test/orb.jl
timholy: I don't understand the need for the `limit` version. If we just return the number of passing tests, can't the caller impose the limit directly? It just seems to complicate the API.

mronian: I was of the opinion that the other version is redundant 😅 My reasons for the `limit` version were (I wanted it to be something like the `isapprox` function)- 

- The other version just returns the mean `diff` over two sets of keypoints. The `mean` could be large if just some outlier pairs of matched keypoints have a very large distance (could be possible that two completely wrong keypoints get matched)
- How do we define passing tests? IMO these would be those pairs which have distance within some `limit`. Also, the mean does not tell us what *proportion* of the keypoints are well matched. 
- The `mean` can mean (pun not intended 😛) different things for differently sized images. For a very large image, a `mean` distance of 15 pixels might mean that the keypoints are quite well matched, whereas for a small one it means they are poorly matched. Instead, if we just pass the `limit` (which can be passed as 1% of the width for example) and get the percentage of keypoints within this limit, it gives a much better way to judge the quality of keypoints obtained without needing to specify the exact distance each time.
- We could return the number of keypoints under the `limit` instead of percentage but that might lead to wrong conclusions (100 keypoints passed looks good but that depends if the total was 110 or 1000).

I might be completely wrong about this though. Does this make sense?

timholy: That's reasonable. I think I was misreading the function and thought it was returning the fraction of wrong tests (but that wouldn't make sense with 1.5...). Carry on :smile:.

mronian: @annimesh2809 lets remove the `mean` part and keep just one version of the function with `limit` and `diff` as the distance measure. The `mean` version is not very useful.

annimesh2809:
`lbp.jl` has been fixed and some manually calculated simple test cases have been added to `test/lbp.jl`.
[Here](https://coveralls.io/github/annimesh2809/ImageFeatures.jl?branch=Test_coveralls) is the coverage report. I have commented out most of the other tests but tests of lbp.jl show 100% coverage.
Regarding the `limit` version of `grade_matches`, should I remove that method with `limit` parameter and create a local method in `test/orb.jl` for the same purpose with a different name so that it does not affect the api?

mronian:
> Regarding the limit version of grade_matches, should I remove that method with limit parameter and create a local method in test/orb.jl for the same purpose with a different name so that it does not affect the api?

Refer [this](https://github.com/JuliaImages/ImageFeatures.jl/pull/25#discussion_r105083255) discussion.

mronian:
@annimesh2809 which other tests have been commented out?

annimesh2809:
@mronian By commented out test cases I meant that while finding coverage I manually commented out other test cases so that travis does not take too much time for building. We only required the coverage of lbp.jl, so I commented out others which took too much time.

mronian:
Oh, thats fine. Once you finish the change to `grade_matches` remember to uncomment the tests.

docs/src/tutorials/brief.md
```diff
@@ -8,7 +8,7 @@ In ImageFeatures.jl we have five methods to determine the vectors `X` and `Y` :
 - [`gaussian`](@ref) : `X` and `Y` are randomly sampled using a Gaussian distribution, meaning that locations that are closer to the center of the patch are preferred
 - [`gaussian_local`](@ref) : `X` and `Y` are randomly sampled using a Gaussian distribution where first `X` is sampled with a standard deviation of `0.04*S^2` and then the `Yi’s` are sampled using a Gaussian distribution – Each `Yi` is sampled with mean `Xi` and standard deviation of `0.01 * S^2`
 - [`random_coarse`](@ref) : `X` and `Y` are randomly sampled from discrete location of a coarse polar grid
-- [`centered`](@ref) : For each `i`, `Xi` is `(0, 0)` and `Yi` takes all possible values on a coarse polar grid
+- [`centre_sample`](@ref) : For each `i`, `Xi` is `(0, 0)` and `Yi` takes all possible values on a coarse polar grid
```
Evizero: Why "cent**re**_sample" and not "cent**er**_sample" ?

I ask because there is a function in `CoordinateTransformations` called `recenter` that we will likely use in `ImageTransformations`. Would be nice to have a consistent spelling.

annimesh2809: @Evizero You are right. For consistency in the API it would be better to call it `center_sample` and not `centre_sample` . I will change it.

mronian:
Are the tests passing now? @annimesh2809

annimesh2809:
@mronian  Yes, all the tests are passing now :smile: .

mronian:
It seems the Travis build failed 🤔

timholy:
`.travis.yml` is only testing `nightly`, should also be testing `0.5`. `appveyor.yml` is testing `0.4` and `nightly`, should be `0.5` and `nightly`.

We expect failures on nightly, due to StaticArrays->ImageFiltering->Images dependency chain. I think everything else in the Images world is ready for 0.6 (but it's hard to test until StaticArrays is ready).

annimesh2809:
The build will likely also fail in 0.5 as the tests for orb.jl take too much time (>45 minutes , travis has per job time limit = 45 mins) with `coverage = true`. But with `coverage = false` all tests pass suceessfully.
You can test it using `Pkg.test("ImageFeatures")` in v0.5

mronian:
@annimesh2809 While we are at it, we should update the REQUIRE and tests to the julia-0.5 style. 

This PR in JuliaImages/ImageDraw.jl#6 is a good reference.

timholy:
I can confirm that this passes tests locally (I'm a big fan of the "view command line instructions" right next to the "Merge pull request" button). I'd like to urge that we merge this soon, since it's the most important of all the queued PRs, and I'm comfortable rebasing my own PR if needed.

@mronian, I'll let you make the decision. You probably already know this, but GitHub recently added the ability to squash before merging (see the pulldown next to "Merge pull request"), so we don't need to trouble @annimesh2809 to do that manually.

timholy:
Given #26, I don't think there's any reason to update that in this PR; we can just merge this and then the others.

mronian:
Sure! I'll do the honours 😄

mronian:
Thanks @annimesh2809! This PR was much needed. Nice work 😄

juliohm:
@mronian , completely off-topic question... what is the emoji code for the thinking face? I googled for minutes and couldn't find the code that GitHub accepts.

Evizero:
did you try typing `:thinking` ?

juliohm:
Great, I should have used github's auto completion features : ) thanks

mronian:
@juliohm LOL 😆

PR #26, opened 2017-03-10
Update Travis and AppVeyor

mronian:


Evizero:
maybe remove 0.5 pre-release support while you are at it. That is something often pointed out at metadata.

https://github.com/JuliaImages/ImageFeatures.jl/blob/master/REQUIRE#L1

mronian:
Tests fail because of the two `centered` functions. This is fixed in #25.

timholy:
I'm going to merge this and then rebase my own PR. Only the 3 together might pass tests!

PR #27, opened 2017-03-10
Fix non-leaftype in ORB and turn off coverage in tests

timholy:
I was trying to see if I could improve the performance of the ORB tests with `Pkg.test("ImageFeatures", coverage=true)`, which really comes down to improving performance when running julia with `julia --check-bounds=yes --inline=no`. Sadly, this doesn't seem to be easy, but at least I found a way to speed up ORB substantially when running under "normal" conditions.

So (after adding the performance improvement) I just turned off coverage in the tests. I think it's more important to have tests than to measure coverage; we can do that locally if we have to.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,15 +10,15 @@ notifications:
 script:
     - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
     - julia -e 'Pkg.add("Distributions")'  # Needed for Docs and Tests
-    - julia -e 'Pkg.add("TestImages")'  # Needed for Docs 
+    - julia -e 'Pkg.add("TestImages")'  # Needed for Docs
     - julia -e 'Pkg.add("ImageMagick")'  # Needed for Docs
-    - julia -e 'Pkg.add("Images")'  # Needed for Docs 
-    - julia -e 'Pkg.clone("https://github.com/JuliaImages/ImageDraw.jl")' # Needed for Docs 
+    - julia -e 'Pkg.add("Images")'  # Needed for Docs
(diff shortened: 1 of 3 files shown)


timholy:
OK, for 0.5 we're now passing on every platform. @mronian, I'll let you make the call about whether it's acceptable to turn off coverage-analysis.

mronian:
Can we run the tests in parallel for Julia ([Reference](https://docs.travis-ci.com/user/speeding-up-the-build/))?

timholy:
Possibly, but we should consider alternatives too:
- use smaller images
- compute the gaussian pyramid once and reuse it. (When inlining is off, that's the first big bottleneck. I didn't have the patience to learn whether there were others later in the process.)
- switch from gaussian pyramid to a `restrict`-pyramid (much less computationally intensive)

If none of these are palatable, then perhaps we should consider a parallel test. Is it worth 45min of computational resources just to measure coverage on each PR?

Evizero:
> Is it worth 45min of computational resources just to measure coverage on each PR?

Over at LossFunctions.jl our travis takes about 1 hrs 30 min to finish the whole test run. It hasn't yet been a practical problem. usually OSX is backed up half the day anyway, so chances are you have to wait for a couple hours for the OSX tests to execute in the first place. For example the full time was 2 hrs 30 min, which means 1 hour was spend waiting in the queue

timholy:
There's a good solution to this problem coming: in Julia 0.6, it `coverage=true` doesn't turn off inlining.

I'd like to propose that we merge this until we can get tests passing on 0.6?

mronian:
Sure 😄

timholy:
Great, now at least CI should work for future PRs.

PR #28, opened 2017-03-15
Added Hough Line Transform

annimesh2809:
Added a new feature detection : Hough line transform

It returns an array of tuples containing (r, theta) as parametric values of lines present in the binary input image.


src/houghtransform.jl
```diff
@@ -0,0 +1,102 @@
+
+#function to compute local maximum lines with values > threshold and return a vector containing them
+function findlocalmaxima(accumulator_matrix::Array{Integer,2},threshold::Integer)
```
timholy: `findlocalmaxima` is already defined in Images.jl. Ideally we would modify `Images.findlocalextrema` to take a function as a first argument, and then you could say `findlocalmaxima(x->x > threshold, accumulator_matrix)`.

src/houghtransform.jl
```diff
@@ -0,0 +1,102 @@
...
+    validLines = Vector{CartesianIndex}(0)
+    for val in CartesianRange(size(accumulator_matrix))
+        if  accumulator_matrix[val] > threshold &&
+            accumulator_matrix[val] > accumulator_matrix[val[1],val[2] - 1] &&
```
timholy: Won't this run into problems on the boundary? Note that `findlocalextrema` already handles boundaries gracefully.

src/houghtransform.jl
```diff
@@ -0,0 +1,102 @@
...
+    rho         = discrete steps for radius
+    theta       = discrete steps for theta
+    min_theta   = minimum theta of lines
+    max_theta   = maximum theta of lines
+    threshold   = no of points to pass through line for considering it valid
+    linesMax    = maximum no of lines to return
+
+"""
+
+function hough_transform_standard{T<:Union{Bool,Gray{Bool}}}(
```
timholy: Would it be a nicer interface to pass in a `Range` for `θ`, e.g. `linspace(min_theta, max_theta, theta)`? That reduces the number of arguments needed for this function. You can get `θ` with `\theta TAB` (no space in between).

src/houghtransform.jl
```diff
@@ -0,0 +1,102 @@
...
+    height = size(img)[1]
+    irho = 1 / rho;
+    numangle::Int = round((max_theta - min_theta)/theta)
+    numrho::Int = round(((width + height)*2 + 1)/rho)
+
+    accumulator_matrix = Array{Integer}(numangle + 2, numrho + 2)
+    fill!(accumulator_matrix, 0)
+
+    #Pre-Computed sines and cosines in tables
+    tabsin = Vector{Number}(numangle)
```
timholy: If you pass in a `Range` for `θ`, then this can become `sinθ, cosθ = sin.(θ).*irho, cos.(θ).*irho`.

src/houghtransform.jl
```diff
@@ -0,0 +1,102 @@
...
+        tabcos[i] = (cos(min_theta + (i-1)*theta) * irho)
+    end
+
+
+    #Hough Transform implementation
+    constadd::Int = round((numrho -1)/2)
+    for pix in CartesianRange(size(img))
+        if img[pix]
+            for i in 1:numangle
+                dist::Int = round((pix[1] * tabsin[i] + pix[2] * tabcos[i]))
```
timholy: You can also use `round(Int, x)` to convert to an `Int`. You also can remove one pair of ().

src/houghtransform.jl
```diff
@@ -0,0 +1,102 @@
...
+Returns an vector of tuples corresponding to the tuples of (r,t) where r and t are parameters for normal form of line:
+    x*cos(t) + y*sin(t) = r
+
+r = length of perpendicular from (0,0) to the line
+t = angle between perpendicular from (0,0) to the line and axis
+
+The lines are generated by applying hough transform on the image.
+
+Parameters:
+    image       = image to apply hough transform on
```
timholy: would be good to mention that it needs to have Bool eltype

src/houghtransform.jl
```diff
@@ -0,0 +1,102 @@
...
+    sort!(validLines, by = (x)->accumulator_matrix[x], rev = true)
+
+    linesMax = min(linesMax, length(validLines))
+
+    lines = Vector{Tuple{Number,Number}}(0)
+
+    #Getting lines with Maximum value in accumulator_matrix && size(lines) < linesMax
+    for l in 1:linesMax
+        lrho = ((validLines[l][2]-1) - (numrho - 1)*0.5)*rho
+        langle = ((validLines[l][1]-1)*theta + min_theta)
```
timholy: Here, you could just look up the corresponding index in `θ`.

test/houghtransform.jl
```diff
@@ -0,0 +1,39 @@
...
+@testset "Hough_Transform" begin
+    @testset "Hough Line Transform" begin
+
+    #For images containing a straight line parallel to axes
+        img = zeros(Bool,10,10)
+        for i in 1:size(img)[1]
+            for j in 1:size(img)[2]
+                img[i,j] = true
+            end
+            h = hough_transform_standard(img,1,0.1,0,3.14,9,2)
```
timholy: Julia defines the constants `pi` and `π` (\pi TAB).

timholy (commented):
Sorry this took so long for me to get to. It looks quite good. The main comment would be that we need to avoid breaking `findlocalmaxima`. You could either define an internal function or extend the one in Images.jl (in Images.jl).

src/houghtransform.jl
annimesh2809: `accumulator_matrix` has size `(numangle + 2, numrho + 2)`, basically applying a padding of width 1 over the entire image. While [filling the accumulator matrix](https://github.com/annimesh2809/ImageFeatures.jl/blob/d17d674a54b3f4b3f066e92e23e4b2456f30ae6a/src/houghtransform.jl#L76), we fill it at `(i+1,dist+1)` rather than `(i,dist)` which prevents any `BoundsError`.
img

timholy: I see. If you really require custom behavior, it's fine to have a private `findlocalmaxima` but obviously don't export it or import it from Images.

src/houghtransform.jl
annimesh2809: Yes, that would indeed be better.

annimesh2809:
I have made the `findlocalmaxima` function private as it required some custom behavior. Also now `θ` is a `Range` object.

Original:
org

After applying `canny`:
canny

Applied `hough_transform_standard` and used `line` to draw the top 7 lines on the image:
hough

src/houghtransform.jl
```diff
@@ -0,0 +1,88 @@
...
+
+r = length of perpendicular from (1,1) to the line
+t = angle between perpendicular from (1,1) to the line and x-axis
+
+The lines are generated by applying hough transform on the image.
+
+Parameters:
+    image       = Image to apply hough transform on (eltype should be `Bool`)
+    ρ           = Discrete step size for perpendicular length of line
+    θ           = Discrete steps for angle of line (should be `range` object)
```
timholy: List of angles `t` for which the transform is computed

src/houghtransform.jl
```diff
@@ -0,0 +1,88 @@
...
+Returns an vector of tuples corresponding to the tuples of (r,t) where r and t are parameters for normal form of line:
+    x*cos(t) + y*sin(t) = r
+
+r = length of perpendicular from (1,1) to the line
+t = angle between perpendicular from (1,1) to the line and x-axis
+
+The lines are generated by applying hough transform on the image.
+
+Parameters:
+    image       = Image to apply hough transform on (eltype should be `Bool`)
```
timholy: Would it be clearer to say "Image to be transformed (eltype should be `Bool`)"?

timholy (commented):
Looks great. If you don't mind making a couple of docstring edits, I think this is ready to go.

timholy:
Thanks very much!

timholy:
Looks like I failed to notice that the docstring doesn't print as formatted in the source code. Can you take a look at it? One (of many) pages describing Markdown format is [here](https://daringfireball.net/projects/markdown/syntax).

annimesh2809:
Yes, some formatting needs to be done. I will get right on it.

PR #29, opened 2017-03-21
added hough circle function

tejus-gupta:
Function for detecting circles in an grayscale image. Unlike the standard circle hough transform, first edges are computed and each edge pixel only votes for circle centers on the normal to the local gradient. For each suspected circle center, radius histogram is produced and the maxima is chosen is the radius. This has smaller space requirements compared to standard hough transform since we don't a 3d array. OpenCV uses the same algorithm for houghcircles.
The function returns two vectors corresponding to the circle centers and their radii. 

I haven't added tests yet.

### DIFF
--- a/src/houghtransform.jl
+++ b/src/houghtransform.jl
@@ -102,7 +102,20 @@ Parameters:
 -   `scale`        = relative accumulator resolution factor  
 -   `min_dist`     = minimum distance between detected circle centers  
 -   `vote_thres`   = accumulator threshold for circle detection  
--   `min_radius:max_radius`   = circle radius range  
+-   `min_radius:max_radius`   = circle radius range
+
+[`canny`](@ref) and [`phase`](@ref) can be used for obtaining img_edges and img_phase respectively.
+
+# Example
+```julia
+img = load("circle.png")
+
+img_edges = canny(img, 1, 0.99, 0.97)
+dx, dy=imgradients(img, KernelFactors.ando5)
+img_phase = phase(dx, dy)
+
+centers, radii=hough_circle_gradient(img_edges, img_phase, 1, 60, 60, 3:50)
+```
 """  
 
 function hough_circle_gradient{T<:Number}(
--- a/src/houghtransform.jl
+++ b/src/houghtransform.jl
@@ -1,4 +1,4 @@
-
+using Images
 """
 ```
 lines = hough_transform_standard(image, ρ, θ, threshold, linesMax)
@@ -106,75 +106,49 @@ Parameters:
     max_radius   = maximum circle radius
 """
 
-function hough_circle_gradient{T<:Integer}(
-        img::AbstractArray{T,2},
+function hough_circle_gradient{N<:Bool, T<:Number}(
+        img_edges::AbstractArray{N,2}, img_phase::AbstractArray{T,2},
         scale::Number, min_dist::Number,
-        canny_thres::Number, vote_thres::Number,
-        min_radius::Integer, max_radius::Integer)
+        vote_thres::Number, radii::AbstractVector{Int})
 
-    rows,cols=size(img)
-
-    img_edges = canny(img, 1, canny_thres, canny_thres/4, percentile=false)
-    dx, dy=imgradients(img, KernelFactors.ando3)
-    _, img_phase = magnitude_phase(dx, dy)
+    rows,cols=size(img_edges)
 
     non_zeros=CartesianIndex{2}[]
     centers=CartesianIndex{2}[]
     circle_centers=CartesianIndex{2}[]
-    circle_radius=Integer[]
-    votes=zeros(Integer, Int(floor(rows/scale))+1, Int(floor(cols/scale))+1)
+    circle_radius=Int[]
+    votes=zeros(Int, Int(floor(rows/scale))+1, Int(floor(cols/scale))+1)
 
     f = CartesianIndex(map(r->first(r), indices(votes)))
     l = CartesianIndex(map(r->last(r), indices(votes)))
 
-    function vote!(img, x, y)
-        i=Int(floor(x))+1
-        j=Int(floor(y))+1
-        p=CartesianIndex(i,j)
+    function vote!(votes, x, y)
+        fx = Int(floor(x))
+        fy = Int(floor(y))
 
-        if min(f,p)==f && max(l,p)==l
-            votes[p]+=1
-        end
-
-        i=Int(floor(x))+2
-        j=Int(floor(y))+1
-        p=CartesianIndex(i,j)
-
-        if min(f,p)==f && max(l,p)==l
-            votes[p]+=1
-        end
-
-        i=Int(floor(x))+1
-        j=Int(floor(y))+2
-        p=CartesianIndex(i,j)
-
-        if min(f,p)==f && max(l,p)==l
-            votes[p]+=1
-        end
-
-        i=Int(floor(x))+2
-        j=Int(floor(y))+2
-        p=CartesianIndex(i,j)
-
-        if min(f,p)==f && max(l,p)==l
-            votes[p]+=1
+        for i in fx:fx+1
+            for j in fy:fy+1
+                if checkbounds(Bool, votes, i, j)
+                    @inbounds votes[i, j] += 1
+                end
+            end
         end
     end
 
-    for i in indices(img, 1)
-        for j in indices(img, 2)
+    for j in indices(img_edges, 2)
+        for i in indices(img_edges, 1)
             if img_edges[i,j]!=0
                 sin_theta = -cos(img_phase[i,j]);
                 cos_theta = sin(img_phase[i,j]);
 
-                for r in min_radius:max_radius
+                for r in radii
                     x=(i+r*sin_theta)/scale
                     y=(j+r*cos_theta)/scale
-                    vote!(img, x, y)
+                    vote!(votes, x, y)
 
                     x=(i-r*sin_theta)/scale
                     y=(j-r*cos_theta)/scale
-                    vote!(img, x, y)
+                    vote!(votes, x, y)
                 end
                 push!(non_zeros, CartesianIndex{2}(i,j));
             end
@@ -190,7 +164,7 @@ function hough_circle_gradient{T<:Integer}(
     sort!(centers, lt=(a, b) -> votes[a]>votes[b])
 
     dist(a, b) = sqrt(sum(abs2, (a-b).I))
-    votes=Array(Integer, Int(floor(dist(f,l))+1))
+    votes=Array(Int, Int(floor(dist(f,l))+1))
 
     for center in centers
         center=(center-1)*scale
@@ -203,7 +177,7 @@ function hough_circle_gradient{T<:Integer}(
                 break
             end
         end
-        if too_close==true
+        if too_close
             continue;
         end
 
--- a/src/houghtransform.jl
+++ b/src/houghtransform.jl
@@ -163,7 +163,7 @@ function hough_circle_gradient{T<:Number}(
 
     f = CartesianIndex(map(r->first(r), indices(votes)))
     l = CartesianIndex(map(r->last(r), indices(votes)))
-    radius_votes=Vector{Int}(Int(floor(dist(f,l))+1))
+    radius_votes=Vector{Int}(Int(floor(dist(f,l)/scale)+1))
 
     for center in centers
         center=(center-1)*scale
@@ -181,11 +181,14 @@ function hough_circle_gradient{T<:Number}(
         end
 
         for point in non_zeros
-            radius_votes[Int(floor(dist(center, point)/scale))+1]+=1
+            r=Int(floor(dist(center, point)/scale))
+            if radii.start/scale<=r<=radii.stop/scale
+                radius_votes[r+1]+=1
+            end
         end
 
         voters, radius = findmax(radius_votes)
-        radius-=1
+        radius=(radius-1)*scale;
 
         if voters>vote_thres
             push!(circle_centers, center)
--- a/src/houghtransform.jl
+++ b/src/houghtransform.jl
@@ -118,9 +118,6 @@ function hough_circle_gradient{T<:Number}(
     circle_radius=Int[]
     votes=zeros(Int, Int(floor(rows/scale))+1, Int(floor(cols/scale))+1)
 
-    f = CartesianIndex(map(r->first(r), indices(votes)))
-    l = CartesianIndex(map(r->last(r), indices(votes)))
-
     function vote!(votes, x, y)
         fx = Int(floor(x))
         fy = Int(floor(y))
--- a/src/houghtransform.jl
+++ b/src/houghtransform.jl
@@ -94,7 +94,7 @@ circle_centers, circle_radius = hough_circle_gradient(img_edges, img_phase, scal
 Returns two vectors, corresponding to circle centers and radius.  
   
 The circles are generated using a hough transform variant in which a non-zero point only votes for circle  
-centers perpendicular to the local gradient.  
+centers perpendicular to the local gradient. In case of concentric circles, only the largest circle is detected.
   
 Parameters:  
 -   `img_edges`    = edges of the image  
--- a/src/houghtransform.jl
+++ b/src/houghtransform.jl
@@ -160,6 +160,9 @@ function hough_circle_gradient{T<:Number}(
     sort!(centers, lt=(a, b) -> votes[a]>votes[b])
 
     dist(a, b) = sqrt(sum(abs2, (a-b).I))
+
+    f = CartesianIndex(map(r->first(r), indices(votes)))
+    l = CartesianIndex(map(r->last(r), indices(votes)))
     radius_votes=Vector{Int}(Int(floor(dist(f,l))+1))
 
     for center in centers
@@ -191,3 +194,4 @@ function hough_circle_gradient{T<:Number}(
     end
     return circle_centers, circle_radius
 end
+
--- a/src/houghtransform.jl
+++ b/src/houghtransform.jl
@@ -157,7 +157,9 @@ function hough_circle_gradient{T<:Number}(
         end
     end
 
-    sort!(centers, lt=(a, b) -> votes[a]>votes[b])
+    @noinline sort_by_votes(centers, votes) = sort!(centers, lt=(a, b) -> votes[a]>votes[b])
+
+    sort_by_votes(centers, votes)
 
     dist(a, b) = sqrt(sum(abs2, (a-b).I))
 
--- a/src/houghtransform.jl
+++ b/src/houghtransform.jl
@@ -137,7 +137,7 @@ function hough_circle_gradient{T<:Number}(
 
     for j in indices(img_edges, 2)::AbstractUnitRange{Int}
         for i in indices(img_edges, 1)::AbstractUnitRange{Int}
-            if img_edges[i,j]!=0
+            if img_edges[i,j]
                 sin_theta = -cos(img_phase[i,j]);
                 cos_theta = sin(img_phase[i,j]);
 
@@ -164,7 +164,7 @@ function hough_circle_gradient{T<:Number}(
     sort!(centers, lt=(a, b) -> votes[a]>votes[b])
 
     dist(a, b) = sqrt(sum(abs2, (a-b).I))
-    radius_votes=Array(Int, Int(floor(dist(f,l))+1))
+    radius_votes=Vector{Int}(Int(floor(dist(f,l))+1))
 
     for center in centers
         center=(center-1)*scale
--- a/src/houghtransform.jl
+++ b/src/houghtransform.jl
@@ -106,8 +106,8 @@ Parameters:
     max_radius   = maximum circle radius
 """
 
-function hough_circle_gradient{N<:Bool, T<:Number}(
-        img_edges::AbstractArray{N,2}, img_phase::AbstractArray{T,2},
+function hough_circle_gradient{T<:Number}(
+        img_edges::AbstractArray{Bool,2}, img_phase::AbstractArray{T,2},
         scale::Number, min_dist::Number,
         vote_thres::Number, radii::AbstractVector{Int})
 
--- a/src/houghtransform.jl
+++ b/src/houghtransform.jl
@@ -134,8 +134,8 @@ function hough_circle_gradient{T<:Number}(
         end
     end
 
-    for j in indices(img_edges, 2)::AbstractUnitRange{Int}
-        for i in indices(img_edges, 1)::AbstractUnitRange{Int}
+    for j in indices(img_edges, 2)
+        for i in indices(img_edges, 1)
             if img_edges[i,j]
                 sinθ = -cos(img_phase[i,j]);
                 cosθ = sin(img_phase[i,j]);
--- a/src/houghtransform.jl
+++ b/src/houghtransform.jl
@@ -127,6 +127,40 @@ function hough_circle_gradient{T<:Integer}(
     f = CartesianIndex(map(r->first(r), indices(votes)))
     l = CartesianIndex(map(r->last(r), indices(votes)))
 
+    function vote!(img, x, y)
+        i=Int(floor(x))+1
+        j=Int(floor(y))+1
+        p=CartesianIndex(i,j)
+
+        if min(f,p)==f && max(l,p)==l
+            votes[p]+=1
+        end
+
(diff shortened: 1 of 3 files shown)


timholy:
Looks like there is both `houghcircle.jl` and `houghcircles.jl`?

tejus-gupta:
Corrected it.

src/houghcircle.jl
```diff
@@ -0,0 +1,134 @@
...
+
+    img_edges = canny(img, 1, canny_thres, canny_thres/4, percentile=false)
+    dx, dy=imgradients(img, KernelFactors.ando3)
+    _, img_phase = magnitude_phase(dx, dy)
+
+    non_zeros=CartesianIndex{2}[]
+    centers=CartesianIndex{2}[]
+    circle_centers=CartesianIndex{2}[]
+    circle_radius=Integer[]
+    votes=zeros(Integer, Int(floor(rows/scale))+1, Int(floor(cols/scale))+1)
```
timholy: `Integer` is an abstract type, and that will kill performance. Use `Int` or, if you need it to be flexible, pass in the concrete type as an argument. (This applies to both `circle_radius` and `votes`.)

src/houghcircle.jl
```diff
@@ -0,0 +1,134 @@
...
+
+    f = CartesianIndex(map(r->first(r), indices(votes)))
+    l = CartesianIndex(map(r->last(r), indices(votes)))
+
+    function vote!(img, x, y)
```
timholy: Misleading because `vote!` doesn't modify `img`, it modifies `votes`. Since it doesn't seem to use `img` at all, just make this `vote!(votes, x, y)`.

src/houghcircle.jl
```diff
@@ -0,0 +1,134 @@
...
+        i=Int(floor(x))+1
+        j=Int(floor(y))+1
+        p=CartesianIndex(i,j)
+
+        if min(f,p)==f && max(l,p)==l
+            votes[p]+=1
+        end
```
timholy: I think this whole "block" can be simplified to
```julia
i = Int(floor(x)) + 1
j = Int(floor(y)) + 1
if checkbounds(Bool, votes, i, j)
    @inbounds votes[i, j] += 1
end
```
Then you don't need `f` and `l`. To me this seems to more clearly convey what's happening here.

src/houghcircle.jl
```diff
@@ -0,0 +1,134 @@
...
+
+        i=Int(floor(x))+2
```
timholy: Do you have to compute `floor` again? Why not have `fx = Int(floor(x))` at the beginning of this function?

src/houghcircle.jl
```diff
@@ -0,0 +1,134 @@
...
+    img          = image to detect circles in
+    scale        = relative accumulator resolution factor
+    min_dist     = minimum distance between detected circle centers
+    canny_thres  = upper threshold for canny, lower threshold=upper threshold/4
+    vote_thres   = accumulator threshold for circle detection
+    min_radius   = minimum circle radius
+    max_radius   = maximum circle radius
+"""
+
+function hough_circle_gradient{T<:Integer}(
```
timholy: Why does it have to be an `Integer` image? Both `canny` and `imgradients` accept anything that's `NumberLike` (number or grayscale).

src/houghcircle.jl
```diff
@@ -0,0 +1,134 @@
...
+        j=Int(floor(y))+2
+        p=CartesianIndex(i,j)
+
+        if min(f,p)==f && max(l,p)==l
+            votes[p]+=1
+        end
+    end
+
+    for i in indices(img, 1)
+        for j in indices(img, 2)
```
timholy: Switch the order of iteration for better performance (`j` in the outer loop, `i` in the inner loop). Julia matrices are stored in column-major order. See https://julialang.org/blog/2013/09/fast-numeric.

src/houghcircle.jl
```diff
@@ -0,0 +1,134 @@
...
+    for i in findlocalmaxima(votes)
+        if votes[i]>vote_thres
+            push!(centers, i);
+        end
+    end
+
+    sort!(centers, lt=(a, b) -> votes[a]>votes[b])
+
+    dist(a, b) = sqrt(sum(abs2, (a-b).I))
+    votes=Array(Integer, Int(floor(dist(f,l))+1))
```
timholy: Using the same name for a completely different object is confusing, and because it has a different type (`Vector` rather than `Matrix`) it makes the algorithm type-unstable. (Also, the concern about `Integer` being an abstract type applies here too.)

src/houghcircle.jl
```diff
@@ -0,0 +1,134 @@
...
+        fill!(votes, 0)
+
+        too_close=false
+        for circle_center in circle_centers
+            if dist(center, circle_center)< min_dist
+                too_close=true
+                break
+            end
+        end
+        if too_close==true
```
timholy: `if too_close` (you don't have to compare it to `true`)

src/houghcircle.jl
```diff
@@ -0,0 +1,134 @@
...
+        canny_thres::Number, vote_thres::Number,
+        min_radius::Integer, max_radius::Integer)
+
+    rows,cols=size(img)
+
+    img_edges = canny(img, 1, canny_thres, canny_thres/4, percentile=false)
+    dx, dy=imgradients(img, KernelFactors.ando3)
+    _, img_phase = magnitude_phase(dx, dy)
+
+    non_zeros=CartesianIndex{2}[]
```
timholy: In general I'd say that allocating things closer to the point where they will first be used makes the code easier to understand. But this is a very minor stylistic point, and if you disagree then just leave it here.

src/houghcircle.jl
```diff
@@ -0,0 +1,134 @@
...
+                    x=(i-r*sin_theta)/scale
+                    y=(j-r*cos_theta)/scale
+                    vote!(img, x, y)
+                end
+                push!(non_zeros, CartesianIndex{2}(i,j));
+            end
+        end
+    end
+
+    for i in findlocalmaxima(votes)
```
timholy: Hmm, this makes me want to write this as
```julia
for i in CartesianRange(indices(votes))
    if islocalmaximum(votes, i)
        push!(centers, i)
    end
end
```
But this is fine for now.

src/houghcircle.jl
```diff
@@ -0,0 +1,134 @@
...
+        end
+        if too_close==true
+            continue;
+        end
+
+        for point in non_zeros
+            votes[Int(floor(dist(center, point)/scale))+1]+=1
+        end
+
+        voters, radius = findmax(votes)
```
timholy: If I understand correctly, this will mean that if you have concentric circles, only one will be detected?

test/houghcircle.jl
```diff
@@ -0,0 +1,26 @@
...
+
+@testset "Hough Circle" begin
+    @testset "Hough Circle Gradient" begin
+
+    dist(a, b) = sqrt(sum(abs2, (a-b).I))
+
+    img=zeros(Integer, 300, 300)
+    for i in CartesianRange(size(img))
+        if dist(i, CartesianIndex(100, 100))<25 || dist(i, CartesianIndex(200, 200))<50
+            img[i]=255
```
timholy: http://juliaimages.github.io/latest/quickstart.html#Colors,-the-0-to-1-intensity-scale,-and-views-1

src/houghcircle.jl
```diff
@@ -0,0 +1,134 @@
...
+centers perpendicular to the local gradient.
+
+Parameters:
+    img          = image to detect circles in
+    scale        = relative accumulator resolution factor
+    min_dist     = minimum distance between detected circle centers
+    canny_thres  = upper threshold for canny, lower threshold=upper threshold/4
+    vote_thres   = accumulator threshold for circle detection
+    min_radius   = minimum circle radius
+    max_radius   = maximum circle radius
```
timholy: You could cut down on the number of parameters by using an argument `radii::AbstractVector` and have the user call this as `hough_circle_gradient(..., min_radius:max_radius)`.

timholy (commented):
Definitely a nice addition.

One overall concern is the API. As you may be gathering, I'm a big proponent of "orthogonal design." Here, the biggest violation of this principle is the fact that you can supply some but not all parameters of `canny` and `imgradients`. What if the choice of `sigma = 1` doesn't work for my image, but `sigma = 2` does? Or I get better orientation information from `ando5`? Then I'm "locked out" of a complex 110-line function for the sake of 3 lines at its beginning.

Now, one solution is to pass all the parameters for both `canny` and `imgradients`. This thought should immediately make you queasy and go scrambling for a better solution. My strategy is to try to reduce the information needed to its bare essentials. Here, notice that the algorithm doesn't actually need `img`: it just needs `img_edges` and `img_phase`. So your best bet is to make the "long function" take those as inputs. 

You can (at your option) also provide a "convenience version" that has the same API as what you've defined here, but all that does is calculate `img_edges` and `img_phase` for you from `img` and then call the "long function." Both versions should be documented so people know about their existence. But note that the "convenience version" isn't really much more convenient; in particular, you still need two inputs, `img` and `canny_thresh`, and users are likely to play around with explicitly calling `canny` directly anyway to tweak `canny_thresh`. So you could consider only defining the version that takes `img_edges` and `img_phase` directly and always place the computation of these quantities in the user's hands. I'm fine with either choice.

An alternative design would be to pass `edgefinder` and `phasecalculator` as function arguments (and have a default `edgefinder = img->canny(img, ...)`), but in this particular case I don't think that's as nice a solution.

annimesh2809:
@tejus-gupta I was wondering if it is necessary to add this functionality in a seperate file `houghcircle.jl`. The file `houghtransform.jl` could contain both `hough_standard_transform` (for the standard line transform) and `hough_circle_gradient` (for hough circles) . Writing separate files for similar features will make the module look cumbersome. @timholy  What do you think?

timholy:
Yes, that seems like a good idea. @tejus-gupta, one other issue is that this is going to need a rebase. It doesn't look as bad as I feared it would be, but if you need any help with that (oh, how I remember my initial painful struggles with git...), don't hesitate to ask!

tejus-gupta:
julia

Why is s() so much slower that than t()?

timholy:
> Why is s() so much slower that than t()?

It's because `rows` and `cols` are non-const global variables. See the very first [performance tip](http://docs.julialang.org/en/latest/manual/performance-tips.html#man-performance-tips-1). Also, a very nice tool is `@code_warntype` (also mentioned on that page):

image

The red highlighting usually indicates a problem, esp. when it's one of the variables highlighted in red. (There are occasional false alarms, though.)

For accurate results on fast-running functions, I recommend BenchmarkTools:
```julia
julia> using BenchmarkTools

julia> @benchmark s() seconds=1
BenchmarkTools.Trial: 
  memory estimate:  2.76 MiB
  allocs estimate:  90601
  --------------
  minimum time:     6.438 ms (0.00% GC)
  median time:      6.501 ms (0.00% GC)
  mean time:        7.099 ms (2.50% GC)
  maximum time:     14.075 ms (13.36% GC)
  --------------
  samples:          141
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%

julia> @benchmark t() seconds=1
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     3.037 ns (0.00% GC)
  median time:      3.041 ns (0.00% GC)
  mean time:        3.056 ns (0.00% GC)
  maximum time:     19.090 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000
  time tolerance:   5.00%
  memory tolerance: 1.00%
```
Its documentation is worth a read, especially [this section](https://github.com/JuliaCI/BenchmarkTools.jl/blob/master/doc/manual.md#interpolating-values-into-benchmark-expressions).

But in this case we get the greatest insight from `@code_llvm`. I won't paste the results for `s()` here (it's quite long), but for `t()` we get
```julia
julia> @code_llvm t()

define void @julia_t_70784() #0 {
top:
  ret void
}
```
You don't have to know much about LLVM IR to understand that this function is trivial. What happened is that LLVM was able to figure out that your `t` function doesn't do any work in those loops, so it just gets rid of them. If you really want to benchmark something like this, make sure your loop does something. This can, in fact, be surprisingly tricky:
```julia
julia> function t2()
           s = 0
           for i in 1:300
               for j in 1:300
                   s += i+j
               end
           end
           s
       end
t2 (generic function with 1 method)

julia> @code_llvm t2()

define i64 @julia_t2_71255() #0 {
top:
  ret i64 27090000
}
```
LLVM realizes that your loops can be reduced to a single computation! So let's try passing in the row/col values as arguments:
```julia
julia> function t3(rows, cols)
           s = 0
           for i in 1:rows
               for j in 1:cols
                   s += i+j
               end
           end
           s
       end
t3 (generic function with 1 method)
```
To make sure you learn how to use these tools, I encourage you to inspect `t3` yourself with `@code_warntype`, `@code_llvm`, and try `@benchmark t3(300,300)`.

tejus-gupta:
Thanks for the help. I have made some changes and the performance has increased significantly :).

tejus-gupta:
Should I rename votes to accumulator_matrix (and same change for radius_votes) to make it consistent with hough_transform_standard's variable names?

tejus-gupta:
There are a couple of issues with this algorithm -   
1. Using sobel derivative to compute the local tangents is only an approximation and the results aren't as accurate as the more memory expensive version with a 3-dimensional accumulator matrix. [OpenCV](http://docs.opencv.org/2.4/modules/imgproc/doc/feature_detection.html?highlight=houghcircles#houghcircles) explicitly warns that the algorithm may give inaccurate radius.

2. As you pointed out, for concentric circle only the largest radius circle is detected. Maybe I should choose all radius above a certain threshold..

src/houghtransform.jl
```diff
@@ -86,3 +86,112 @@ function hough_transform_standard{T<:Union{Bool,Gray{Bool}}}(
     lines
 
 end
+
+"""
+```
+circle_centers, circle_radius = hough_circle_gradient(img, scale, min_dist, canny_thres, vote_thres, min_radius, max_radius)
+```
```
timholy: Docstring needs to be updated to the new argument. 

Note also the formatting issue as described in other comments. Because it looks so nice in the source code, I am very likely to fail to notice such problems when I review code (sorry). To make sure I don't accidentally merge something with formatting problems, by far your best option is to try all of your docstrings yourself, e.g., `?hough_circle_gradient` at the REPL, to make sure it looks correct.

src/houghtransform.jl
```diff
@@ -86,3 +86,112 @@ function hough_transform_standard{T<:Union{Bool,Gray{Bool}}}(
...
+
+    f = CartesianIndex(map(r->first(r), indices(votes)))
+    l = CartesianIndex(map(r->last(r), indices(votes)))
+
+    function vote!(votes, x, y)
+        fx = Int(floor(x))
+        fy = Int(floor(y))
+
+        for i in fx:fx+1
+            for j in fy:fy+1
```
timholy, line 139: :+1:

src/houghtransform.jl
```diff
@@ -86,3 +86,112 @@ function hough_transform_standard{T<:Union{Bool,Gray{Bool}}}(
...
+                if checkbounds(Bool, votes, i, j)
+                    @inbounds votes[i, j] += 1
+                end
+            end
+        end
+    end
+
+    for j in indices(img_edges, 2)::AbstractUnitRange{Int}
```
timholy: You shouldn't need the type annotation. Did it show some kind of instability when you checked with `@code_warntype`?

src/houghtransform.jl
```diff
@@ -86,3 +86,112 @@ function hough_transform_standard{T<:Union{Bool,Gray{Bool}}}(
...
+        for i in indices(img_edges, 1)::AbstractUnitRange{Int}
+            if img_edges[i,j]!=0
```
timholy: Since `eltype(img_edges)` is `Bool`, this can just be `if img_edges[i,j]`.

src/houghtransform.jl
```diff
@@ -86,3 +86,112 @@ function hough_transform_standard{T<:Union{Bool,Gray{Bool}}}(
...
+    img          = image to detect circles in
+    scale        = relative accumulator resolution factor
+    min_dist     = minimum distance between detected circle centers
+    canny_thres  = upper threshold for canny, lower threshold=upper threshold/4
+    vote_thres   = accumulator threshold for circle detection
+    min_radius   = minimum circle radius
+    max_radius   = maximum circle radius
+"""
+
+function hough_circle_gradient{N<:Bool, T<:Number}(
```
timholy: `Bool` doesn't have any subtypes (it's a concrete type, or "leaftype": `isleaftype(Bool) -> true`). So you can get rid of the `N` parameter and just use `img_edges::AbstractMatrix{Bool}`.

src/houghtransform.jl
```diff
@@ -86,3 +86,112 @@ function hough_transform_standard{T<:Union{Bool,Gray{Bool}}}(
...
+                    @inbounds votes[i, j] += 1
+                end
+            end
+        end
+    end
+
+    for j in indices(img_edges, 2)::AbstractUnitRange{Int}
+        for i in indices(img_edges, 1)::AbstractUnitRange{Int}
+            if img_edges[i,j]!=0
+                sin_theta = -cos(img_phase[i,j]);
```
timholy: I do find `sinθ` more attractive. If you use the REPL or [several editors](https://github.com/JuliaEditorSupport), you can get that by typing `sin\theta` and then hitting Tab.

This comment is not merge-blocking :smile:, I'll be happy to merge it with the current names.

src/houghtransform.jl
```diff
@@ -86,3 +86,112 @@ function hough_transform_standard{T<:Union{Bool,Gray{Bool}}}(
...
+    for i in findlocalmaxima(votes)
+        if votes[i]>vote_thres
+            push!(centers, i);
+        end
+    end
+
+    sort!(centers, lt=(a, b) -> votes[a]>votes[b])
+
+    dist(a, b) = sqrt(sum(abs2, (a-b).I))
+    radius_votes=Array(Int, Int(floor(dist(f,l))+1))
```
timholy: If you're changing other stuff, might as well make this
```julia
radius_votes = Vector{Int}(Int(floor(dist(f,l))+1))
```
(The version you have causes a deprecation warning on 0.6; we'll just have to fix that anyway.)

test/houghtransform.jl
```diff
@@ -36,4 +36,30 @@ using ImageFeatures
         er = sum(map((t1,t2) -> abs(t1-t2), theta, [0, π/2, π/2, 0]))
         @test er <= 0.1
     end
+
+    @testset "Hough Circle Gradient" begin
+
+    dist(a, b) = sqrt(sum(abs2, (a-b).I))
+
+    img=zeros(Integer, 300, 300)
```
timholy: `zeros(Int, 300, 300)`. About the only place you should use `Integer` is `arg::Integer` in function signature declarations.

test/houghtransform.jl
```diff
@@ -36,4 +36,30 @@ using ImageFeatures
...
+    for i in CartesianRange(size(img))
+        if dist(i, CartesianIndex(100, 100))<25 || dist(i, CartesianIndex(200, 200))<50
+            img[i]=1
+        else
+            img[i]=0
+        end
+    end
+
+    img_edges = canny(img, 1, 0.2, 0.1, percentile=false)
```
timholy, line 53: Obviously this will need to be updated once a new Images gets tagged, but that needs to wait on https://github.com/JuliaImages/Images.jl/pull/617.

timholy (commented):
Looking pretty close. I started this review before you added some changes, so some comments may already be outdated.

timholy:
> Should I rename votes to accumulator_matrix (and same change for radius_votes) to make it consistent with hough_transform_standard's variable names?

That's a good idea to make them the same. I don't have strong feelings about whether they should be `votes` or `accumulator_matrix`, I'm fine with changing either one.

> Using sobel derivative to compute the local tangents is only an approximation and the results aren't as accurate as the more memory expensive version with a 3-dimensional accumulator matrix. OpenCV explicitly warns that the algorithm may give inaccurate radius.

Out of curiosity, do you do better with `ando5`? This seems like a case where it would be worth spending a little more time on the gradient computation if it makes what would be an `O(N^3)` part faster by using a faster algorithm.

> Maybe I should choose all radius above a certain threshold..

Presumably they should be local maxima in terms of votes. But yes, that seems like a good idea.

src/houghtransform.jl
tejus-gupta: Updated.
```
-   `img_edges`    = edges of the image  
-   `img_phase`    = phase of the gradient image 
``` 
Is the meaning clear from this?

src/houghtransform.jl
tejus-gupta: You are right. I was trying out some performance upgrade tips and forgot to delete this.  
Deleted it now.

src/houghtransform.jl
timholy: Overall yes, but you should direct users to documentation about how they might obtain these arguments. See the use of `@ref` in various places in the Images, documentation, as well as ones that have an example in their docstring (e.g., the documentation for `imcorner` in https://github.com/JuliaImages/Images.jl/pull/617).

tejus-gupta:
```
sort!(centers, lt=(a, b) -> votes[a]>votes[b])
```
I noticed that this was taking a lot of time and almost 60% of total memory used. I think this is because of cost of dereferencing the array each time sort! needs a comparison. So I used a dictionary to store the votes instead.
```
votes_array=Dict{CartesianIndex{2},Int}()

for i in findlocalmaxima(votes)
    if votes[i]>vote_thres
        push!(centers, i);
        votes_array[i]=votes[i]
    end
end

sort!(centers, lt=(a, b) -> votes_array[a]>votes_array[b])
```
before --> after
wohoo

Any cleaner way to do this?

tejus-gupta:
```
sort!(centers, lt=(a, b) -> votes[a]>votes[b])
```
I noticed that this was taking a lot of time and almost 60% of total memory used. I think this is because of cost of dereferencing the array each time sort! needs a comparison. So I used a dictionary to store the votes instead.
```
votes_array=Dict{CartesianIndex{2},Int}()

for i in findlocalmaxima(votes)
    if votes[i]>vote_thres
        push!(centers, i);
        votes_array[i]=votes[i]
    end
end

sort!(centers, lt=(a, b) -> votes_array[a]>votes_array[b])
```
before --> after
wohoo

Nice speedup! :)

Any better way to do this?

timholy:
> `sort!(centers, lt=(a, b) -> votes[a]>votes[b])`

> I noticed that this was taking a lot of time and almost 60% of total memory used.

Nice job noticing this! You're encountering https://github.com/JuliaLang/julia/issues/15276. A workaround is to change the `sort!` call to `sort_by_votes(centers, votes)`, where
```julia
@noinline sort_by_votes(centers, votes) = sort!(centers, lt=(a, b) -> votes[a]>votes[b])
```
That `@noinline` is potentially quite important; to work around issue 15276, you need this to be a separate function call, and if the compiler inlines it then your workaround doesn't work. @annimesh2809, this might be an issue for your "lines" code, too.

If you make that change, with `@code_warntype` you should be able to convince yourself that your function is completely inferrable.

annimesh2809:
Yes, It seems like I overlooked this performance issue, will be careful from next time. I will fix this issue (and also fix improper docstring formatting) in a separate PR. @tejus-gupta Nice work :)

timholy:
These things are hard to notice. If you use the most recent version of ProfileView, type-instabilities should be colored in red; that makes it a lot easier to detect.

tejus-gupta:
For concentric circles, the strategy to simply take all radii corresponding to local maxima above vote_thres in radius histogram wouldn't work straightaway. 
rad
The radius histogram is a bit noisy and so there are some false peaks. 10% non-maximal suppression around the peaks worked for me across a wide variety of images.
Should I add this as a parameter or simply use 10% always?

The only other library to use this variant of hough transform, OpenCV simply takes the global maxima of the radius histogram and doesn't work for concentric circles.

timholy:
The 10% seems like it can't be a universal, so I'm reluctant to hard-code that.

One option would be to document that it doesn't handle concentric circles. The other option might be to pass the "peak selector" as an anonymous function, with some kind of default choice for users who want that (e.g., using 10% non-maximal suppression).

tejus-gupta:
I am in favor of documenting that it doesn't handle concentric circles.

timholy:
Thanks, @tejus-gupta!

PR #30, opened 2017-03-22
Add Distributions.jl requirement

mprat:
When doing `Pkg.clone("https://github.com/JuliaImages/ImageFeatures.jl")` on my local machine in Julia 0.5 I discovered that Distributions is a required package.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,9 +1,5 @@
 julia 0.5
+ColorTypes
 Images 0.5
-Colors 0.6
-ColorVectorSpace 0.1
 Distributions 0.12
 FixedPointNumbers 0.1
-FileIO
-Compat 0.7.15
-StatsBase
--- a/REQUIRE
+++ b/REQUIRE
@@ -2,6 +2,7 @@ julia 0.5
 Images 0.5
 Colors 0.6
 ColorVectorSpace 0.1
+Distributions 0.12
 FixedPointNumbers 0.1
 FileIO
 Compat 0.7.15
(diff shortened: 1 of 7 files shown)


Evizero:
Good catch. It looks to me like there are mutlitple things going on with the REQUIRE file. For example `FileIO` looks out of place. Could be worth rethinking what exactly to put there

Evizero:
While you are at it, mind removing `Distributions` from `test/REQUIRE`, since then it is no longer needed there and duplicate information.

The other possibly redundant REQUIRE entries are a subject for another time

mprat:
Ok, I updated REQUIRE, test/REQUIRE, and the .travis.yml to avoid redundant package installs. I wish there were a docs/REQUIRE standard too, but that's a separate issue.

Evizero:
lgtm. Improves the current situation. @timholy or @mronian can one of you take a final look and merge?

REQUIRE
```diff
@@ -1,8 +1,5 @@
 julia 0.5
+ColorTypes
 Images 0.5
-Colors 0.6
-ColorVectorSpace 0.1
+Distributions 0.12
 FixedPointNumbers 0.1
```
timholy, line 5: We should also bump the FPN min version to 0.3; it now uses `Normed`, which is the new name in 0.3.

timholy (commented):
Very helpful indeed. If you're willing to make one more small change, that would be a bonus, but this could be merged in its current form.

timholy:
You know, I think I'm being ungrateful here. Let's merge this and I'll take care of the FPN version. Thanks, @mprat!

mprat:
No problem at all. I'm new to the Julia ecosystem so I am hesitant to bump package versions I don't know really well. Gives me a good excuse to look at FPN next!

PR #31, opened 2017-03-27
Refactor tests to use Base.Test instead of FactCheck

Evizero:


timholy:
I like it.

This will be a source of (and magnet for) merge conflicts, so I suspect we should merge this soon. Any reservations? CC @annimesh2809 (you can be sure to be informed of all discussion here by clicking "watch" in the upper right corner).

Evizero:
Functionality wise this PR is finished, so feel free to merge or close at your discretion. 

I don't currently use ImageFeatures, so I have little opinion on this. Generally speaking it seems to be a good idea to me to use a common testing scheme throughout the Images ecosystem.

Evizero:
I just realize there is a mix of tabs and spaces, some of which I introduced somehow (probably had paste mode on in VIM which bypasses the replace-tabs-with-spaces settings). We should fix that before merging.

timholy:
That would be fine. You probably noticed I fixed a merge conflict. Looks like CI is going to pass on 0.5 and fail on 0.6; once you fix the tabs feel free to merge.

Evizero:
Actually there were so many trailing whitespaces and tabs to replace, that I'd feel more comfortable if you'd approve it again before I merge

timholy (approved):

annimesh2809:
@timholy I wasn't aware of the `notifications` and `watch` options on GitHub and was introduced to it recently :sweat_smile: . I didn't see the notification of this PR until now and so did not respond to your reply.  This PR was much required.

timholy:
Great, glad we didn't mess anything up that you were working on.

PR #34, opened 2017-04-29
Performance fix for hough_transform_standard

annimesh2809:
A performance fix for `hough_transform_standard` function by taking into account [JuliaLang/julia#15276](https://github.com/JuliaLang/julia/issues/15276). I have also fixed the docstring, now it displays properly.

Performance before:
before_pf

Performance after:
after_pf

`@code_warntype` confirms that everything is type-stable now.


annimesh2809:
Test errors are occurring because Images.jl v0.9.1 is being used instead of [v0.10.0](https://github.com/JuliaImages/Images.jl/releases/tag/v0.10.0), hence the `Percentile` functionality cannot be used. Should I revert the [test of hough circles ](https://github.com/JuliaImages/ImageFeatures.jl/pull/34/files#diff-5e74cbfbe61b2c628f3865ea97f29641R53)to its previous form?

timholy:
If we can merge https://github.com/JuliaLang/METADATA.jl/pull/9043 relatively soon, we can go back to making Images great again :smile:. CC @tkelman. So I'd hold off for a day or two?

tkelman:
working on 0.5.2 but keep getting interrupted by people breaking things on 0.6. pr should be up in a day or two, release another couple days after that if testing is clean

src/houghtransform.jl
```diff
@@ -1,84 +1,100 @@
...
     end
 
     ρ > 0 || error("Discrete step size must be positive")
 
-    height, width = size(img)
+    height, width = indices(img)
     ρinv = 1 / ρ
     numangle = length(θ)
-    numrho = round(Integer,(2(width + height) + 1)*ρinv)
+    numrho = round(Int,(2(last(width) + last(height)) + 1)*ρinv)
```
timholy: Maybe `length(width)` rather than `last(width)`? (The two are the same only if the indices start with 1, so it's necessary to pick one.)

src/houghtransform.jl
```diff
@@ -1,84 +1,100 @@
...
                 push!(validLines,val)
             end
         end
-        validLines
     end
 
     ρ > 0 || error("Discrete step size must be positive")
 
-    height, width = size(img)
+    height, width = indices(img)
```
timholy: Might be slightly odd to continue to use these names if these are ranges rather than numbers.

src/houghtransform.jl
```diff
@@ -1,84 +1,100 @@
...
 
     linesMax = min(linesMax, length(validLines))
 
-    lines = Vector{Tuple{Number,Number}}(0)
+    lines = Vector{Tuple{Float64,Float64}}(0)
 
     #Getting lines with Maximum value in accumulator_matrix && size(lines) < linesMax
-    for l in 1:linesMax
-        lrho = ((validLines[l][2]-1) - (numrho - 1)*0.5)*ρ
+    for l in range(1,linesMax)
```
timholy: Any reason for this change? I think `1:linesMax` is more explicit, if the two are equivalent.

timholy (commented):
Testing is working out now, sorry for the delay here.

src/houghtransform.jl
annimesh2809: I was trying different methods and forgot to revert back. I will change that.

PR #35, opened 2017-04-30
Censure Detector

tejus-gupta:
Added StarFilter to #1

### DIFF
--- a/src/censure.jl
+++ b/src/censure.jl
@@ -240,6 +240,153 @@ function _filter_response(int_imgs::Tuple, OF::OctagonFilter)
     
 end
 
+function _filter_response(int_imgs::Tuple, SF::StarFilter)
+    int_img = int_imgs[1]
+    rs_img = int_imgs[2]
+    ls_img = int_imgs[3]
+
+    margin = SF.a_out + Int(floor(SF.a_out/2))
+    img_shape = size(int_img)
+    response = zeros(T, img_shape)
+    R = CartesianRange(CartesianIndex((margin + 2, margin + 2)), CartesianIndex((img_shape[1] - margin, img_shape[2] - margin))) 
+
+    a_in2=Int(floor(SF.a_in/2))
+    a_out2=Int(floor(SF.a_out/2))
+    
+    for I in R
+        #inner star
+        in_sum=0
+
+        topleft = I + CartesianIndex(-SF.a_in - a_in2 - 1, -1)
+        topright = I + CartesianIndex(-SF.a_in - a_in2 - 1, 0)
+        bottomleft = I + CartesianIndex(-SF.a_in - 1, -a_in2)
+        bottomright = I + CartesianIndex(-SF.a_in - 1, a_in2 - 1)
+        A = checkbounds(Bool, int_img, topleft) ? ls_img[topleft] : zero(T)
+        B = checkbounds(Bool, int_img, topright) ? rs_img[topright] : zero(T)
+        C = checkbounds(Bool, int_img, bottomleft) ? rs_img[bottomleft] : zero(T)
+        D = checkbounds(Bool, int_img, bottomright) ? ls_img[bottomright] : zero(T)
+        in_sum += B + D - A - C
+
+        topleft = I + CartesianIndex(-SF.a_in - 1, -SF.a_in - 1)
+        topright = I + CartesianIndex(-SF.a_in - 1, SF.a_in)
+        bottomleft = I + CartesianIndex(-a_in2 , -SF.a_in - 1)
+        bottomright = I + CartesianIndex(-a_in2 , SF.a_in)
+        A = checkbounds(Bool, int_img, topleft) ? int_img[topleft] : zero(T)
+        B = checkbounds(Bool, int_img, topright) ? int_img[topright] : zero(T)
+        C = checkbounds(Bool, int_img, bottomleft) ? int_img[bottomleft] : zero(T)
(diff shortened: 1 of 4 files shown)


src/censure.jl
```diff
@@ -0,0 +1,420 @@
...
+    out_weight :: Float64
+
+end
+
+type CENSURE{F} <: Params
+
+    smallest :: Int
+    largest :: Int
+    filter_type :: Type{F}
+    filter_stack :: Array{F}
```
timholy, line 44: Specify dimensionality or make it a parameter, `Array{F,N}`. Otherwise it's an abstract type and will deliver slow performance.

src/censure.jl
```diff
@@ -0,0 +1,420 @@
...
+    in_weight :: Float64
+    out_weight :: Float64
+
+end
+
+type CENSURE{F} <: Params
+
+    smallest :: Int
+    largest :: Int
+    filter_type :: Type{F}
```
timholy, line 43: I don't think this needs to be a field. You could define
```julia
eltype{F}(::Type{CENSURE{F}}) = F
eltype(censure::CENSURE) = eltype(typeof(censure))
```

src/censure.jl
```diff
@@ -0,0 +1,420 @@
...
+    CENSURE{filter}(smallest, largest, filter, _get_filter_stack(filter, smallest, largest), response_threshold, line_threshold)
+end
+
+function censure{T, F}(img::AbstractArray{T, 2}, params::CENSURE{F})
+    int_img = _get_integral_image(img, params.filter_stack[1])
+    responses = map(f -> _filter_response(int_img, f), params.filter_stack)
+    response_matrix = reshape(hcat(responses...), size(img)..., size(responses)...)
+    minima, maxima = extrema_filter(convert(Array{Float64}, padarray(response_matrix, [1, 1, 1], [1, 1, 1], "replicate")), [3, 3, 3])
+    features = map(i -> (minima[i] == response_matrix[i] || maxima[i] == response_matrix[i]) && ( response_matrix[i] > params.response_threshold ), CartesianRange(size(response_matrix)))
+    (grad_x, grad_y) = imgradients(img, "sobel", "replicate")
```
timholy, line 400: Presumably deprecated

src/censure.jl
```diff
@@ -0,0 +1,420 @@
...
+
+function CENSURE(; smallest::Integer = 1, largest::Integer = 7, filter::Type = BoxFilter, response_threshold::Number = 0.15, line_threshold::Number = 10)
+    CENSURE{filter}(smallest, largest, filter, _get_filter_stack(filter, smallest, largest), response_threshold, line_threshold)
+end
+
+function censure{T, F}(img::AbstractArray{T, 2}, params::CENSURE{F})
+    int_img = _get_integral_image(img, params.filter_stack[1])
+    responses = map(f -> _filter_response(int_img, f), params.filter_stack)
+    response_matrix = reshape(hcat(responses...), size(img)..., size(responses)...)
+    minima, maxima = extrema_filter(convert(Array{Float64}, padarray(response_matrix, [1, 1, 1], [1, 1, 1], "replicate")), [3, 3, 3])
```
timholy, line 398: I think this is deprecated

src/censure.jl
```diff
@@ -0,0 +1,420 @@
...
+    features = map(i -> (minima[i] == response_matrix[i] || maxima[i] == response_matrix[i]) && ( response_matrix[i] > params.response_threshold ), CartesianRange(size(response_matrix)))
+    (grad_x, grad_y) = imgradients(img, "sobel", "replicate")
+    cov_xx = grad_x .* grad_x
+    cov_xy = grad_x .* grad_y
+    cov_yy = grad_y .* grad_y
+    for i in 1:params.largest - params.smallest + 1
+        gamma = (1 + (params.smallest + i - 1) / 3.0)
+        filt_cov_xx = imfilter_gaussian(cov_xx, [gamma, gamma])
```
timholy, line 406: Deprecated

src/censure.jl
```diff
@@ -0,0 +1,420 @@
...
+        filt_cov_xy = imfilter_gaussian(cov_xy, [gamma, gamma])
+        filt_cov_yy = imfilter_gaussian(cov_yy, [gamma, gamma])
+
+        features[:, :, i] = map((xx, yy, xy, f) -> (xx + yy) ^ 2 > params.line_threshold * (xx * yy - xy ^ 2) ? false : f, filt_cov_xx, filt_cov_yy, filt_cov_xy, features[:, :, i])
+    end
+    keypoints = Array{Keypoint}([])
```
timholy, line 412: Abstract type? `keypoints = Vector{Keypoint}(0)` or `keypoints = Keypoint[]`.

src/censure.jl
```diff
@@ -0,0 +1,420 @@
...
+    scales = Array{Integer}([])
```
timholy, line 413: Here too.

test/censure.jl
```diff
@@ -0,0 +1,110 @@
...
+		@test isapprox(response[5, 5], 1.35417, rtol = 0.001) == true
+		@test all(response[:, 1:3] .== 0) == true 
+		@test all(response[1:3, :] .== 0) == true
+		@test all(response[:, 6:6] .== 0) == true 
+		@test all(response[6:7, :] .== 0) == true 
+
+
+	end
+
+end
```
timholy, line 110: Will users call `censure` directly? If so there needs to be a test. We should make sure that all filter types are tested, too.

timholy (changes requested):
This is much more in @mronian's area of expertise, but it's looking promising to me.

PR #37, opened 2017-07-27
Added Histogram of Gradient Features

tejus-gupta:
Added function to compute HOG descriptor according to this paper - [Histograms of Oriented Gradients for Human Detection](https://hal.inria.fr/file/index/docid/548512/filename/hog_cvpr2005.pdf). I have set the default parameters to the best parameters for pedestrian detection as reported in the paper.

Will add tests by tomorrow.

### DIFF
--- a/src/hog.jl
+++ b/src/hog.jl
@@ -3,6 +3,8 @@
 hog_params = HOG([orientations = 9], [cell_size = 8], [block_size = 2], [block_stride = 1], [norm_method = "L2-norm"])
 ```
 
+Histogram of Oriented Gradient (HOG) is a dense feature desciptor usually used for object detection. See "Histograms of Oriented Gradients for Human Detection" by Dalal and Triggs.
+
 Parameters:  
 -    orientations   = number of orientation bins
 -    cell_size      = size of a cell is cell_size x cell_size (in pixels)
--- a/src/hog.jl
+++ b/src/hog.jl
@@ -112,7 +112,7 @@ function create_hog_descriptor{T<:Images.NumberLike}(mag::AbstractArray{T, 2}, p
 
             y1 = (bin_y1-1)*cell_size+cell_size/2
             y2 = (bin_y2-1)*cell_size+cell_size/2
-            b_y = abs(i[2]-y1)+abs(y2-i[2])
+            b_y = cell_size
 
             hist[bin_θ1, bin_x, bin_y1] += w*(1-(phase[i]-θ1)/b_θ)*(1-(i[2]-y1)/b_y)
             hist[bin_θ1, bin_x, bin_y2] += w*(1-(phase[i]-θ1)/b_θ)*(1-(y2-i[2])/b_y)
@@ -127,7 +127,7 @@ function create_hog_descriptor{T<:Images.NumberLike}(mag::AbstractArray{T, 2}, p
 
             x1 = (bin_x1-1)*cell_size+cell_size/2
             x2 = (bin_x2-1)*cell_size+cell_size/2
-            b_x = abs(i[1]-x1)+abs(x2-i[1])
+            b_x = cell_size
 
             hist[bin_θ1, bin_x1, bin_y] += w*(1-(phase[i]-θ1)/b_θ)*(1-(i[1]-x1)/b_x)
             hist[bin_θ1, bin_x2, bin_y] += w*(1-(phase[i]-θ1)/b_θ)*(1-(x2-i[1])/b_x)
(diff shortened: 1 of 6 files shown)


timholy:
I've gotten more than one email recently from folks asking about adding HOG, so clearly there's a burning need for this :smile:.

src/hog.jl
```diff
@@ -0,0 +1,99 @@
...
+    block_size::Int
+    block_stride::Int 
+    norm_method::String
+end
+
+function HOG(; orientations::Int = 9, cell_size::Int = 8, block_size::Int = 2, block_stride::Int = 1, norm_method::String = "L2-norm")
+    HOG(orientations, cell_size, block_size, block_stride, norm_method)
+end
+
+function create_descriptor{T<:Images.NumberLike}(img::AbstractArray{T, 2}, params::HOG)
```
timholy: Possible to do color too? I recognize that HOG may be difficult to generalize to >2d, so I'm fine with keeping the restriction to 2d.

src/hog.jl
```diff
@@ -0,0 +1,99 @@
+"""
+```
+hog_params = HOG([orientations = 9], [cell_size = 8], [block_size = 2], [block_stride = 1], [norm_method = "L2-norm"])
+```
```
timholy, line 4: Would be good to have a brief description of what this is for, before going into detail about the parameters.

src/hog.jl
```diff
@@ -0,0 +1,99 @@
...
+
+    for i in R
+        lower = floor(Int, phase[i]*orientations/180) + 1
+        upper = lower%orientations + 1
+
+        cell_i::Int = floor((i[1] - 1)/cell_rows) + 1
+        cell_j::Int = floor((i[2] - 1)/cell_cols) + 1
+
+        #votes are weighted by gradient magnitude and linearly interpolated between neighboring bin centers
+        if lower != 9
```
timholy: Should this be `lower <= orientations`?

timholy (commented):
Looks promising. I'll be interested in seeing what profiling turns up as the bottleneck(s), it's possible there are some areas for improvement.

src/hog.jl
kmsquire: There are various extensions of HOG to color.  The original Dalal and Triggs paper suggests computing the HOG in each channel separately, and choosing the value from the channel with the highest gradient magnitude.

tejus-gupta: Thanks for pointing it out to me. I read the paper once more and found that I had missed one more detail - The votes need to be interpolated in orientation as well as position i.e. a pixel will not only vote for two orientation bins in its cell, but also in the 4 neighboring cells.

timholy:
Just "request a review" when you're ready for another round.

tejus-gupta:
@timholy I have updated the PR with support for color images and trilinear interpolation for histograms. I have added tests that compute the descriptor for grayscale/color images and verify that descriptor length is correct. I wanted to add correctness tests but didn't know how to get the "correct" vector. Scikit-image's implementation is slightly different and doesn't use interpolation for computing histogram. [Here](https://github.com/opencv/opencv/blob/master/modules/objdetect/src/hog.cpp) is opencv's implementation but I was having difficulty in understanding the code (if histograms are concatenated row-wise or column-wise etc).

(I couldn't find the 'request a review button'.)

tejus-gupta:
Another way to test the correctness would be to train a SVM classifier using HOG descriptors and compare accuracy with the paper's results (which I am anyways going to do for the tutorial).

timholy:
Hmm, yes, testing is a little tricky. While SVM classification would be useful, it's a pretty indirect measure of the effectiveness/accuracy of the individual computations.

It seems that the main complexity lies in the trilinear interpolation. How about splitting that out into a separate function and testing that directly? You have several cases that you can set up to make sure it's working properly.

tejus-gupta:
I have added the tests.

```create_descriptor``` takes avg. 2.713 ms for 128x64 grayscale image and avg. 7.224 ms for 128x64 RGB image.

File: test/runtests.jl [hunk 4]
```diff
@@ -47,15 +47,16 @@ function _reverserotate(p, angle, center)
...
-    "houghtransform.jl",
+    # "core.jl",
+    # "brief.jl",
+    # "glcm.jl",
+    # "lbp.jl",
+    # "corner.jl",
+    # "orb.jl",
+    # "freak.jl",
+    # "brisk.jl",
+    # "houghtransform.jl",
```
timholy: Just need to un-comment these and I think this is good to go.

timholy:
Note the test failures. Also, can you add 0.6 to `.travis.yml` and `appveyor.yml`? I would argue we should also drop support for 0.5.

timholy:
Yay, great to have this in!

tejus-gupta:
It seems like I a made a small error.
```
phase = abs.(phase*180/pi)
θ = phase[i, j]
bin_θ1 = floor(Int, θ*orientations/180) + 1
```
I mistakenly thought that ```phase``` returns value in range (-pi, pi). It returns values in range (-pi, pi] and so when ```phase[i, j]=pi```,  ```θ = 180``` and the third equation becomes incorrect.
```bin_θ1 = min(floor(Int, θ*orientations/180) + 1, orientations)``` should work. I will submit a PR for the correction shortly.

timholy:
Test too, please!

PR #42, opened 2017-08-27
Tutorial - Object Detection using HOG

tejus-gupta:
Added tutorial for object detection using HOG. I have also corrected a bug found in the current implementation.

In the tutorial, I have provided a classification data-set with cropped images from a pedestrian detection data-set. I will have to check if this is permitted. If it is not, I will link to the original dataset and provide scripts to get the dataset in required format.


docs/src/tutorials/object_detection.md
```diff
@@ -0,0 +1,122 @@
+# Object Detection using HOG
+
+In this tutorial, we will use Histogram of Oriented Gradient (HOG) feature descriptor based linear SVM to create a person detector. We will first create
+a person classifier and then use this classifier with a sliding window to identify and localize people in an image.
+
+The key challenge in creating a classifier is that it needs to work with variations in illumination, pose and occlusions in the image. To achieve this, we will train
+the classifier on an intermediate representation of the image instead of the pixel-based representation. Our ideal representation (commonly called feature vector)
+captures information which is useful for classification but is invariant to to small changes in illumination and occlusions. HOG descriptor is a gradient-based
```
timholy: "to to"

docs/src/tutorials/object_detection.md
```diff
@@ -0,0 +1,122 @@
...
+representation which is invariant to local geometric and photometric changes (i.e. shape and illumination changes) and so is a good choice for our problem. In fact
+HOG descriptors are widely used for object detection.
+
+We will use [this data](https://drive.google.com/file/d/0B9V0KF3ZHWtWa0hGNHJDSm1XZnM/view?usp=sharing) for training our classifier. We will start by loading the data and computing HOG features of all the images.
+
+```julia
+using Images, ImageFeatures
+
+pos_examples = "path_to_data/human/"
```
timholy: human->humans

docs/src/tutorials/object_detection.md
```diff
@@ -0,0 +1,122 @@
...
+n = n_pos + n_neg                       # number of training examples 
+data = Array{Float64}(n, 3780)          # Array to store HOG descriptor of each image. Each image in our training data has size 128x64 and so has a 3780 length 
+labels = Vector{Int}(n)                 # Vector to store label (1=human, 0=not human) of each image.
+
+i = 0
+for file in [readdir(pos_examples); readdir(neg_examples)]
+    i += 1
+    filename = "$(i <= n_pos ? pos_examples : neg_examples )$file"
+    img = load(filename)
+    examples[i, :] = create_descriptor(img, HOG())
```
timholy: Do you mean `data`? And if you agree with my transpose above, should be `data[:, i]`.

docs/src/tutorials/object_detection.md
```diff
@@ -0,0 +1,122 @@
...
+train_ind = random_perm[1:2500]
+test_ind = random_perm[2501:end]
+
+model = svmtrain(data[:, train_ind], labels[train_ind]);
+```
+
+Now let's test this classifier on some images.
+
+```julia
+img = load("humans/per00003.ppm")
```
timholy: `load("$pos_examples/per00003.ppm")`

docs/src/tutorials/object_detection.md
```diff
@@ -0,0 +1,122 @@
...
+
+```julia
+using LIBSVM
+
+#Split the dataset into train and test set. Train set = 2500 images, Test set = 122 images.
+random_perm = randperm(n)
+train_ind = random_perm[1:2500]
+test_ind = random_perm[2501:end]
+
+model = svmtrain(data[:, train_ind], labels[train_ind]);
```
timholy, line 45: Noticed that the docstring for `svmtrain` seems wrong. If you agree, did you report this?

docs/src/tutorials/object_detection.md
```diff
@@ -0,0 +1,122 @@
...
+neg_examples = "path_to_data/not_humans/"
+
+n_pos = length(readdir(pos_examples))   # number of positive training examples
+n_neg = length(readdir(neg_examples))   # number of negative training examples
+n = n_pos + n_neg                       # number of training examples 
+data = Array{Float64}(n, 3780)          # Array to store HOG descriptor of each image. Each image in our training data has size 128x64 and so has a 3780 length 
+labels = Vector{Int}(n)                 # Vector to store label (1=human, 0=not human) of each image.
+
+i = 0
+for file in [readdir(pos_examples); readdir(neg_examples)]
```
timholy: You could use `for (i, file) in enumerate([readdir(pos_examples); readdir(neg_examples)])` here. An advantage is that if someone following this has to re-run the loop, `i` won't be out-of-sync if it error partway through on a previous run (and `i` is no longer set to 0 upon entry).

If you do that, get rid of the `i += 1`.

docs/src/tutorials/object_detection.md
```diff
@@ -0,0 +1,122 @@
...
+```julia
+using Images, ImageFeatures
+
+pos_examples = "path_to_data/human/"
+neg_examples = "path_to_data/not_humans/"
+
+n_pos = length(readdir(pos_examples))   # number of positive training examples
+n_neg = length(readdir(neg_examples))   # number of negative training examples
+n = n_pos + n_neg                       # number of training examples 
+data = Array{Float64}(n, 3780)          # Array to store HOG descriptor of each image. Each image in our training data has size 128x64 and so has a 3780 length
```
timholy: I suspect this should be `Array{Float64}(3780, n)` (the transpose of what you have).

docs/src/tutorials/object_detection.md
```diff
@@ -0,0 +1,122 @@
...
+```julia
+img = load("humans/per00003.ppm")
+descriptor = Array{Float64}(3780, 1)
+descriptor[:, 1] = create_descriptor(img, HOG())
+
+predicted_label, _ = svmpredict(model, descriptor);
+print(predicted_label)                          # 1=human, 0=not human
+
+# Get test accuracy of our model
+(predicted_labels, decision_values) = svmpredict(model, data[:, test_ind]);
```
timholy: Don't need the parentheses on the LHS

docs/src/tutorials/object_detection.md
```diff
@@ -0,0 +1,122 @@
...
+| ![Original](../img/human3.png) | ![Original](../img/not-human1.jpg) |
+|:------:|:---:|
+| predicted_label = 1 | predicted_label = 0 |
+
+Next we will use our trained classifier with a sliding window to localize persons in an image.
+
+![Original](../img/humans.jpg)
+
+```julia
+img = load("humans.jpg")
```
timholy: Maybe `load("$pos_examples/../humans.jpg")`? Or tell people to navigate to the data directory and get rid of your `path_to_data` above.

docs/src/tutorials/object_detection.md
```diff
@@ -0,0 +1,122 @@
...
+rows, cols = size(img)
+
+scores = Array{Float64}(22, 46)
```
timholy: The last column never seems to be set here.

docs/src/tutorials/object_detection.md
```diff
@@ -0,0 +1,122 @@
...
+
+scores[scores.>0] = 0
+object_locations = findlocalminima(scores)
+
+rectangles = [[((i[2]-1)*10+1, (i[1]-1)*10+1), ((i[2]-1)*10+64, (i[1]-1)*10+1), ((i[2]-1)*10+64, (i[1]-1)*10+128), ((i[2]-1)*10+1, (i[1]-1)*10+128)] for i in object_locations];
+
+for rec in rectangles
+    draw!(img, Polygon(rec), RGB{N0f8}(0, 0, 1.0))
+end
+imshow(img)
```
timholy, line 111: No package supporting `imshow` has been loaded (e.g., ImageView or ImageInTerminal).

timholy (commented):
:fireworks: A few comments, but otherwise seems ready to merge. Nice work!

Also, thanks in advance for checking the permissions.

docs/src/tutorials/object_detection.md
tejus-gupta: I noticed that too but didn't report the issue.

timholy: Better than reporting...https://github.com/mpastell/LIBSVM.jl/pull/30

src/hog.jl
```diff
@@ -112,7 +112,7 @@ function create_hog_descriptor(mag::AbstractArray{T, 2}, phase::AbstractArray{T,
 end
 
 function trilinear_interpolate!(hist, w, θ, orientations, i, cell_size, cell_rows, cell_cols, rows, cols)
-    bin_θ1 = floor(Int, θ*orientations/180) + 1
+    bin_θ1 = min(floor(Int, θ*orientations/180) + 1, orientations)
```
timholy, line 115: Is there a test that would have caught this bug?

In any event, this seems read to merge.

docs/src/tutorials/object_detection.md
```diff
@@ -6,16 +6,16 @@ a person classifier and then use this classifier with a sliding window to identi
...
 captures information which is useful for classification but is invariant to small changes in illumination and occlusions. HOG descriptor is a gradient-based
-representation which is invariant to local geometric and photometric changes (i.e. shape and illumination changes) and so is a good choice for our problem. In fact
-HOG descriptors are widely used for object detection.
+representation which is invariant to local geometric and photometric changes (i.e. shape and illumination changes) and so is a good choice for our problem. In fact HOG descriptors are widely used for object detection.
 
-We will use [this data](https://drive.google.com/file/d/0B9V0KF3ZHWtWa0hGNHJDSm1XZnM/view?usp=sharing) for training our classifier. We will start by loading the data and computing HOG features of all the images.
+Run the the script get_data.bash provided [here](https://drive.google.com/open?id=0B9V0KF3ZHWtWR1dBR2VZUDctUGc) (Change the variables `path_to_tutorial` in preprocess.jl and path to julia executable in get_data.bash). This script will download the required datasets. We will start by loading the data and computing HOG features of all the images.
```
timholy: Is that Google drive repository likely to be permanent? Maybe also say that the whole thing needs to be downloaded and decompressed, and then run `get_data.bash`.

Too bad that script will only work on Unix, but ATM I don't see a great way to decompress a `.tar.gz` purely within Julia.

timholy (approved):

src/hog.jl
tejus-gupta: Added a test.

docs/src/tutorials/object_detection.md
tejus-gupta: The drive repository should be permanent. I have updated the part and explained that the file needs to be downloaded and decompressed too.

timholy:
I'm fine with you squash+merging this whenever you're ready.

tejus-gupta:
I don't have write access to merge this. I think this is ready to be merged.

timholy:
Check your email :smile:

tejus-gupta:
Thanks for adding me :smile:. I still don't have write access to this repository. You would have to add me  as a maintainer.

timholy:
Hmm, bad settings. Can you try again?

PR #58, opened 2018-10-04
polish hough_circle_gradient

jw3126:
If you like it I can do the same for `hough_transform_standard`.


timholy (approved):
This is lovely, I want this.

If you deliberately trigger the depwarn, do you like it? Should it be a bit more like https://github.com/JuliaImages/ImageFiltering.jl/pull/78?

jw3126:
Yes I think the depwarn is fine. This time its easier since there were no default arguments before.
```julia
julia> hough_circle_gradient(img_edges, img_phase, 1,1,1,1:1)
┌ Warning: `hough_circle_gradient(img_edges, img_phase, scale, min_dist, vote_threshold, radii)` is deprecated, use `hough_circle_gradient(img_edges, img_phase, radii; scale=scale, min_dist=min_dist, vote_threshold=vote_threshold)` instead.
│   caller = top-level scope at none:0
└ @ Core none:0
(CartesianIndex{2}[], Int64[])
```

test/houghtransform.jl
```diff
@@ -1,25 +1,36 @@
...
     #For images containing a straight line parallel to axes
-        img = zeros(Bool,10,10)
-        for i in 1:size(img)[1]
-            for j in 1:size(img)[2]
-                img[i,j] = true
-            end
-            h = hough_transform_standard(img,1,range(0,step=π/2/99,length=100),9,2)
+        for i in 1:9
+            img = zeros(Bool,9,9)
+            img[i,:] = true
```
timholy: `.=`

src/houghtransform.jl
```diff
@@ -116,26 +124,29 @@ Parameters:
...
 ```
 """  
 function hough_circle_gradient(
-        img_edges::AbstractArray{Bool,2}, img_phase::AbstractArray{T,2},
-        scale::Number, min_dist::Number,
-        vote_thres::Number, radii::AbstractVector{Int}) where T<:Number
+        img_edges::AbstractArray{Bool,2},
+        img_phase::AbstractArray{<:Number,2},
+        radii::AbstractVector{<:Integer};
+        scale=1::Number,
```
timholy: You've got the `::` on the wrong component:
```julia
julia> f(; x=1::Int) = true
f (generic function with 1 method)

julia> f(x="hello")
true

julia> g(; x::Int=1) = true
g (generic function with 1 method)

julia> g(x="hello")
ERROR: TypeError: in #g, in typeassert, expected Int64, got String
Stacktrace:
 [1] (::getfield(Main, Symbol("#kw##g")))(::NamedTuple{(:x,),Tuple{String}}, ::typeof(g)) at ./none:0
 [2] top-level scope at none:0
```

src/houghtransform.jl
```diff
@@ -37,13 +43,18 @@ julia> lines = hough_transform_standard(img_edges, 1, linspace(0,π,30), 40, 5)
...
 """  
 function hough_transform_standard(
-    img::AbstractArray{T,2},
-    ρ::Real, θ::AbstractRange,
-    threshold::Integer, linesMax::Integer) where T<:Union{Bool,Gray{Bool}}
+    img::AbstractMatrix;
+    stepsize=1,
+    angles=range(0,stop=pi,length=minimum(size(img))),
+    vote_threshold=minimum(size(img)) / stepsize -1,
+    max_linecount=typemax(Int))
```
timholy, line 51: If you're removing the `eltype` restriction, we need tests on other eltypes (e.g., `Gray{N0f8}`, `RGB{Float32}`).

src/houghtransform.jl
```diff
@@ -116,26 +124,29 @@ Parameters:
 -   `img_phase`    = phase of the gradient image   
 -   `scale`        = relative accumulator resolution factor
```
timholy, line 127: Ensure the parameters are described in the order they appear in the docstring

src/houghtransform.jl
```diff
@@ -116,26 +124,29 @@ Parameters:
...
 ```julia
 img = load("circle.png")
 
-img_edges = canny(img, 1, 0.99, 0.97)
+img_edges = canny(img, (Percentile(99), Percentile(97)))
 dx, dy=imgradients(img, KernelFactors.ando5)
 img_phase = phase(dx, dy)
 
-centers, radii=hough_circle_gradient(img_edges, img_phase, 1, 60, 60, 3:50)
+centers, radii=hough_circle_gradient(img_edges, img_phase, 3:50)
```
timholy: This might yield different results from the original example (`min_dist` and `vote_threshold` of 3 rather than 60). You've checked to make sure they are reasonable?

src/houghtransform.jl
```diff
@@ -81,30 +90,29 @@ function hough_transform_standard(
...
     end
 
     lines
-
 end
 
 """
 ```
-circle_centers, circle_radius = hough_circle_gradient(img_edges, img_phase, scale, min_dist, vote_thres, min_radius:max_radius)  
+circle_centers, circle_radius = hough_circle_gradient(img_edges, img_phase, radii; scale=1, min_dist=minimum(radii), vote_threshold=minimum(radii))
```
timholy: Should the default for `vote_threshold` also depend on `scale`?

timholy (commented):
Looks really nice, thanks for cleaning this up!

src/houghtransform.jl
jw3126: Maybe, but I wanted to keep things simple stupid. The situation is this:
* A perfect circle or radius r will vote r*pi times for its center. It will do this regardless of the scale.
* A real world circle spreads the votes over a neighborhood of the true center
* Depending on the scale parameter, this neighborhood may be conflated to a single point
* If `length(radii) > 1` the circle does multiple voting rounds and votes from nearby radii will accidentally hit the center also. Especially if scale is large.

So a more complicated heuristic would be something like 
```const * minimum(radii) * min(scale, length(radii))```
Would you prefer that? Also is the algorithm in the package published somewhere?

File: src/houghtransform.jl (hunk 5 above)
Comment (jw3126) [reply src/houghtransform.jl]: I must confess, I don't know how to run the example. Where can I get "circle.png" from?

File: src/houghtransform.jl (hunk 3 above)
Comment (jw3126) [reply src/houghtransform.jl:51]: Ah good catch, no `img` should be the edges, e.g. `Bool`.

File: src/houghtransform.jl (hunk 5 above)
Comment (timholy) [reply src/houghtransform.jl]: Good point. No clue, sorry.

Should we demo it on `img = load(download("http://docs.opencv.org/3.1.0/water_coins.jpg"))`? We already use that one [here](https://juliaimages.github.io/latest/imagesegmentation.html#Watershed-1).

File: src/houghtransform.jl (hunk 6 above)
Comment (timholy) [reply src/houghtransform.jl]: Not sure about the publication. Any thoughts, @tejus-gupta?

**Comment by timholy** [comment]
This is fantastic, thanks so much!

File: src/houghtransform.jl (hunk 6 above)
Comment (tejus-gupta) [reply src/houghtransform.jl]: The algorithm is described in [The Adaptive Hough Transform by Illingworth and Kittler](https://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=4767964) (see Section-2 Circle Finding). 

As far as I can tell, a threshold of `minimum(radii)*min(scale, length(radii))` selects circles that have at least `1/(2pi)` fraction of votes compared with a full circle with radius `minimum(radii)`.

Does it make more sense to interpret the `vote_threshold` as the fraction of votes for a circle compared with a full circle of the same radius? Then the threshold for selecting centers would be `vote_threshold * minimum(radii) * min(scale, length(radii))` and the threshold for selecting radius for each center would be `vote_threshold * r * min(scale, length(radii))`. The second threshold would depend on radius `r`.

Comment (jw3126) [reply src/houghtransform.jl]: Thanks for the reference (unfortunatelly I cannot access the paper though)!
I like the idea of providing the vote threshold as a fraction of full circle votes very much.
The api could be to pass `vote_threshold = FullCircleVotes(0.4)`.

I think you are right that in an ideal situation `minimum(radii)*min(scale, length(radii))` corresponds to `1/(2pi)` fraction of votes. But there are complications:
* The dependence of full circle votes on scale etc is very spiky
* Larger circles have more votes then smaller ones
* If the true radius becomes extremal `r_true in extrema(radii)` there are only half as much nearby radius votes, as for `r_true` in the middle of `radii`.

At least the first problem could be overcome by actually counting the votes for a perfect circle:
```julia
function compute_vote_threshold(obj::FullCircleVotes, img, scale, radii...)
     # select some r in the middle of radii
     # create perfect circle of radius r
     # count its vote for its true center
end
```

Issue #48, opened 2018-02-03
HOG not working


davidbp:
Hello,
I have tried the HOG code but it does not seem to work.
The following code:

```julia
img = testimage("lena_gray")
create_descriptor(img, HOG()) 
```
returns

```
BoundsError: attempt to access 9×32×32 Array{Float64,3} at index [10, 1, 1]

Stacktrace:
 [1] trilinear_interpolate!(::Array{Float64,3}, ::Float32, ::Float32, ::Int64, ::CartesianIndex{2}, ::Int64, ::Int64, ::Int64, ::Int64, ::Int64) at /Users/davidbuchaca1/.julia/v0.6/ImageFeatures/src/hog.jl:132
 [2] create_hog_descriptor(::Array{Float32,2}, ::Array{Float32,2}, ::ImageFeatures.HOG) at /Users/davidbuchaca1/.julia/v0.6/ImageFeatures/src/hog.jl:82
 [3] create_descriptor(::Array{ColorTypes.Gray{FixedPointNumbers.Normed{UInt8,8}},2}, ::ImageFeatures.HOG) at /Users/davidbuchaca1/.julia/v0.6/ImageFeatures/src/hog.jl:34
 [4] include_string(::String, ::String) at /Applications/Julia-0.6.app/Contents/Resources/julia/lib/julia/sys.dylib:?
```

Am I doing something wrong?

colbec:
I have the same issue when running the train script on openSUSE 42.3 using the published binary for Julia 0.6. It stops at the first image in the humans directory (./humans/per00001.ppm) with a message similar to the one reported.

`ERROR: LoadError: BoundsError: attempt to access 9×16×8 Array{Float64,3} at index [10, 1, 1]`

zygmuntszpak:
This bug is resolved on the master branch but a release was subsequently never tagged. Hence when you do Pkg.add("ImageFeatures") you currently don't get the latest version. If you checkout the master branch things will work as expected. It looks like _tejus-gupta_ fixed the problem as part of commit #42. @Evizero could you please tag a release when you have a spare moment?

Evizero:
Thanks for the reminder

https://github.com/JuliaLang/METADATA.jl/pull/13878

colbec:
Thanks for this, which fixes that particular issue; now the process stops when trying to process the non_human images, finding an issue with incompatible images, the size of which must be a multiple of the cell size:
```
[...]

Issue #44, opened 2017-11-05
Hog tutorial not working


davidbp:
The tutorial folder here:
http://juliaimages.github.io/ImageFeatures.jl/latest/tutorials/object_detection.html

contains a bash script that is not working since  [url]( http://cbcl.mit.edu/projects/cbcl/software-datasets/pedestrians128x64.tar.gz) is not valid.  Since there is another url that seems to work maybe you could simply remove the first one.

kmsquire:
Thanks for the report.  Do you feel like submitting a pull request?  Since this is a simple change, you could to navigate to https://github.com/JuliaImages/ImageFeatures.jl/blob/master/docs/src/tutorials/object_detection.md and click on the pencil in the upper right to edit the file.

tejus-gupta:
The tutorial uses this pedestrian dataset - http://cbcl.mit.edu/software-datasets/PedestrianData.html. The download link isn't working right now and I can't find dataset anywhere else.

Thanks for reporting this. I will update the tutorial to use another dataset (by this weekend).

davidbp:
I have found the data here:

https://github.com/mbeyeler/opencv-machine-learning/tree/78f5f6ecfbbae3f61b378060d60641420331115c/notebooks/data/chapter6

Maybe you could simply add it in the repo and load it from the repo. When compressed its only some MB.

davidbp:
An even better idea would be to add the data in MLDatasets.jl so other people can use it independently of ImageFeatures easily.

abhijithch:
Also, there might be a small typo in the `get_data.bash` file. The name of the Julia file in the bash file(line number 11) is `preprocess_data.jl` whereas the actual file is `preprocess.jl`. I could have sent a PR, but this is in the downloaded Tutorial.zip folder.

tejus-gupta:
@timholy Should I add the contents of Tutorial.zip to this repository itself? Since its on google drive currently, only I can update the content of the tutorial.

timholy:
I'm a little reluctant to add a 10MB file to the repository (and once added it becomes a bottleneck on the git history even if we delete it later). What about adding a download script to MLDatasets.jl as suggested by @davidbp? It would be a bit like TestImages.jl.

Evizero:
[...]

PR #60, opened 2019-01-06
Fix data download script in HoG Tutorial

Deepank308:
The download link is now fixed for downloading pedestrian dataset as provided by @davidbp in #44.
The odd size of 65x129 is also fixed as reported by @colbec in #48.

### DIFF
--- a/docs/src/tutorials/object_detection.md
+++ b/docs/src/tutorials/object_detection.md
@@ -8,7 +8,7 @@ the classifier on an intermediate representation of the image instead of the pix
 captures information which is useful for classification but is invariant to small changes in illumination and occlusions. HOG descriptor is a gradient-based
 representation which is invariant to local geometric and photometric changes (i.e. shape and illumination changes) and so is a good choice for our problem. In fact HOG descriptors are widely used for object detection.
 
-Download the script to get the training data [here](https://drive.google.com/open?id=0B9V0KF3ZHWtWR1dBR2VZUDctUGc). Download tutorial.zip, decompress it and run get_data.bash. (Change the variable `path_to_tutorial` in preprocess.jl and path to julia executable in get_data.bash). This script will download the required datasets. We will start by loading the data and computing HOG features of all the images.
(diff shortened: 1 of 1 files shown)


timholy:
Thanks!

timholy:
This is broken now. @Deepank308, @davidbp can you resolve this?

davidbp:
I think we end up verifying that the data used is [here](https://github.com/mbeyeler/opencv-machine-learning/blob/78f5f6ecfbbae3f61b378060d60641420331115c/notebooks/data/chapter6/pedestrians128x64.tar.gz)

The data from the MIT web [here](http://cbcl.mit.edu/software-datasets/PedestrianData.htmll) is not working.

I verified that the data from the MIT url has 924 examples of size 64 x 128 matches the data fro the github repo found in `pedestrians128x64.tar.gz`

Would It make sense to add `ImageDatasets.jl` (similar to `MLDatasets.jl`) containing a bunch of datasets for different Image related tasks?

johnnychen94:
> Would It make sense to add ImageDatasets.jl (similar to MLDatasets.jl) containing a bunch of datasets for different Image related tasks?

This could go to TestImages.jl if there's no issue about the license

Unfortunately, the pedestrian images dataset belongs to the CBCL dataset, which doesn't have a free public license.

> Copyright 2000
Center for Biological & Computational Learning at MIT and MIT
All rights reserved.
Permission to copy and modify this data, software, and its documentation only for internal research use in your organization is hereby granted, provided that this notice is retained thereon and on all copies. This data and software should not be distributed to anyone outside of your organization without explicit written authorization by the author(s) and MIT. It should not be used for commercial purposes without specific permission from the authors and MIT. MIT also requires written authorization by the author(s) to publish results obtained with the data or software and possibly citation of relevant CBCL reference papers.
We make no representation as to the suitability and operability of this data or software for any purpose. It is provided "as is" without express or implied warranty.
> http://poggio-lab.mit.edu/codedatasets

I think we need to find some alternative images for these tutorials.

davidbp:
What do you think about doing a face detection equivalent? Any face dataset that is open and we could add to testimages?

Issue #61, opened 2019-03-29
docstring missing


johnnychen94:
The current julia documentation fails to build: https://github.com/JuliaImages/juliaimages.github.io/issues/44#issuecomment-453773548

When building Images.jl Documentation https://github.com/JuliaImages/juliaimages.github.io with `Documenter v0.22`, there'll be lots of unresolved links if using `@docs` macro in [function_reference.md](https://github.com/JuliaImages/juliaimages.github.io/commit/8f910cbec5b2b24b9c55bed95e32275c34994969#diff-0eca5f6512d2cce06ad3a484380dfeb8L383)

```markdown
```@docs
glcm
...
``` #
```

and `glcm` doesn't have its docstring
```julia
help?> glcm
search: glcm glcm_prop glcm_norm glcm_var_ref glcm_entropy glcm_mean_ref glcm_symmetric glcm_var_neighbour

  No documentation found.
```
To supprese the no-doc-found warnings, I temporarily changed `@docs` to `julia` in https://github.com/JuliaImages/juliaimages.github.io/pull/50

The following is a list of undocumented methods

- [ ] glcm
- [ ] glcm_symmetric
- [ ] glcm_norm
- [ ] glcm_prop
- [ ] max_prob
- [ ] contrast
- [ ] ASM
- [ ] IDM
- [ ] glcm_entropy
- [ ] energy
- [ ] dissimilarity
- [ ] correlation
- [ ] glcm_mean_ref
- [ ] glcm_mean_neighbour
- [ ] glcm_var_ref
- [ ] glcm_var_neighbour
- [ ] lbp
- [ ] modified_lbp
- [ ] direction_coded_lbp
- [ ] lbp_original
- [ ] lbp_uniform
- [ ] lbp_rotation_invariant
- [ ] multi_block_lbp

cc: @timholy

Jay-sanjay:
is the issue resolved sir

ashwanirathee:
It's not actually, as you might notice here: https://github.com/JuliaImages/ImageFeatures.jl/blob/master/src/brisk.jl all those functions are documented and we could use help :)

Jay-sanjay:
Sir starting with first function glcm and the related glcm functions just needed to confirm https://github.com/JuliaImages/ImageFeatures.jl/blob/master/docs/src/tutorials/glcm.md?plain=1#L1-L26 
I have to make changes in these .md files  am I right ?

PR #64, opened 2019-04-26
Get the docs building on recent Julia

timholy:


johnnychen94:
the `function_reference.html` page is broken due to missing docstrings https://github.com/JuliaImages/ImageFeatures.jl/issues/61

timholy:
Agreed those need fixing. But since they are just warnings, I'll go ahead and merge this, just so we get updated docs (e.g., `axes` vs `indices`).

timholy:
Deployment indeed worked (check `latest`)

PR #71, opened 2020-01-06
update package versions

johnnychen94:
* Don't directly rely on ColorTypes, Colors and FixedPointNumbers
(https://github.com/JuliaImages/Images.jl/issues/802)
* Package versions incompatible to Julia v1.0 are dropped
* remove ">=" notation in compat section and leave compat maintainance work to
CompatHelper

Images.jl as a dependency is quite large, but for now, I'd like to leave it as it is since we haven't spent much on developing it recently.

cf. This PR blocks https://github.com/JuliaImages/juliaimages.github.io/pull/104

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,28 +1,23 @@
 name = "ImageFeatures"
 uuid = "92ff4b2b-8094-53d3-b29d-97f740f06cef"
-version = "0.3.0"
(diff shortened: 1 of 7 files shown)


johnnychen94:
Hope this makes the last piece to JuliaImages/Images.jl#802

PR #88, opened 2021-02-01
Added_comment_to_docs_code

krishna-vasudev:
Helps in better understanding,and hence avoid confusion for new users.


timholy:
Unfortunately this is a bit misleading, because it mutates the output so won't each element of the vector be identical, each containing all the matches? This seems like a situation where it would be better to use `foreach` than `map`.

If you want to draw just one match, you can just call `draw!` manually.

krishna-vasudev:
> Unfortunately this is a bit misleading, because it mutates the output so won't each element of the vector be identical, each containing all the matches? This seems like a situation where it would be better to use `foreach` than `map`.
> 
> If you want to draw just one match, you can just call `draw!` manually.

I got it.So i am closing this pr.

PR #95, opened 2021-05-29
docs: upgrade dependencies and left align the table

hyrodium:
This PR includes the following updates:

## Fix image generation with `@example block`
In #94, I've added `using Rotations`, but the fix was incomplete and [the documentation](https://juliaimages.org/ImageFeatures.jl/dev/tutorials/brisk/) still doesn't include the generated images.
This was because `docs/Project.toml` doesn't have dependence to `Rotations`, and I've fixed this.

image

(This screenshot is from https://github.com/JuliaImages/ImageFeatures.jl/runs/2695863640?check_suite_focus=true)

## Replace Lena with Lighthouse
In [tutorials/brief section](https://juliaimages.org/ImageFeatures.jl/dev/tutorials/brief/), there was a warning.

>┌ Warning: Usage of "lena" is not recommended, and the image may be removed in a later release. See https://womenlovetech.com/losing-lena-why-we-need-to-remove-one-image-and-end-techs-original-sin/ for more information.
└ @ TestImages ~/.julia/packages/TestImages/mpUUF/src/TestImages.jl:89

I've updated the image to `testimage("lighthouse")`.

image

## Update version of Documenter.jl
`v0.24` to `v0.26`

## Markdown table alignment
Right-aligned
image

to Left-aligned
image

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -9,4 +9,4 @@ Rotations = "6038ab10-8711-5258-84ad-4b1120ba62dc"
 TestImages = "5e47fb64-e119-507b-a336-dd2b206d9990"
 
 [compat]
-Documenter = "0.24"
+Documenter = "0.26"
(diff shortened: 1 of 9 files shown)


johnnychen94 (approved):
Looks better to me, thanks!

PR #102, opened 2023-07-21
Update Project.toml's compat

ashwanirathee:
- It wasn't working with Images 0.25

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,10 +1,15 @@
 name: CI
 on:
-  pull_request:
   push:
     branches:
       - master
     tags: '*'
+  pull_request:
+concurrency:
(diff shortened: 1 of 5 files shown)


ashwani-rathee:
1 test fails

timholy:
xref #103

timholy:
After we merge this, perhaps we should slim the dependencies down to ImageCore and other specific packages (including ImageCorners). That might make it possible to move this into Images someday.

timholy:
This won't pass tests but I have local fixes.

timholy:
BTW I disabled testing on nightly, see https://discourse.julialang.org/t/why-do-packages-run-continuous-integration-tests-on-julia-nightly/101208/3?u=tim.holy

PR #105, opened 2023-07-23
WIP: replace Images with specific packages

timholy:
Currently the  bottleneck is `gaussian_pyramid` which is still in
Images.jl

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,20 +4,26 @@ version = "0.5.0"
 
 [deps]
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
-Images = "916415d5-f1e6-5110-898d-aaa5f9f070e0"
+ImageBase = "c817782e-172a-44cc-b673-b171935fbb9e"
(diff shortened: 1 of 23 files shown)


ashwanirathee:
Can we bring pyramid functions here or somewhere more appropriate if needed?That almost empties out algorithms.jl in Images.jl if threshold ones are gone too today and much closer to a meta package than ever.

timholy:
Pyramids should be a separate package; ImageTracking among likely many others needs them. Though perhaps they could move to ImageBase; it's a pretty fundamental concept in image processing.

ashwanirathee:
Should we make it a separate package then or keep it in ImageBase.jl? I am assuming it doesn't conflict with monorepo work because Images.jl is at top level and ImageBase.jl seems to be separate enough too.

timholy:
I don't have a strong opinion, but I think since `restrict` is in `ImageBase` and we might use `restrict` in making fast pyramids, I think moving it to ImageBase seems fine.

PR #114, opened 2025-03-31
Update freak.md

caesquerre:
The angle in the description (75 degrees) didn't correspond to the one in the code 5pi/6 (150 degrees). The rotated image showed has also a rotation of 150 degrees.


timholy:
Thanks!

PR #115, opened 2025-03-31
Update brief.md

caesquerre:
The lighthouse is used in the example code as well in the examples for ORB, BRISK and FREAK


timholy:
Thanks!

